PR #12, opened 2020-02-05
Ajuste no backgroundColor

claudneysessa:
Ajuste no backgroundColor


jonataslaw:
Added backgroundColor + 23 features on latest version

PR #73, opened 2020-05-06
Create .gitignore

hdeyana:


Nipodemos:
what happened with this PR? i actually have to remove a lot of auto-generated files from my PR everytime because it does not have  a .gitignore to make them go away, so i think that would be a nice idea!!

if you want something more complete (although i think the one from this PR is just fine) you can use the [gitignore creator website](https://www.toptal.com/developers/gitignore), and it will generate one base on what you choose, in this case, dart and flutter

What do you think?

jonataslaw:
I think I missed this PR, I'll merge. 
Thanks

PR #74, opened 2020-05-06
Fix broken code in Snackbar example

paolorotolo:


Nipodemos:
i guess this pull request is no longer required since the changes made in README

jonataslaw:
Thanks for that, I think it has already been corrected in the readme, but thanks for having dedicated your time to this PR.

paolorotolo:
No problem, thanks for your work :)

PR #111, opened 2020-05-17
Added ChangeThemeMode

RodBr:
This change will allow Get to track the current system brightness or a user specified override of the theme.


jonataslaw:
Thanks for this! 
LGTM.

PR #131, opened 2020-05-22
Add unit tests

chimon2000:
This is a very modest attempt to improve the quality of the package by adding widget tests.

My knowledge of the inner workings are very limited, so I was unable to test certain features such as snackbar (which gives odd errors) or rx (which is fairly new) fully.

Here's a summary of the code coverage on the first pass:

image

### DIFF
--- a/test/bottomsheet_test.dart
+++ b/test/bottomsheet_test.dart
@@ -1,0 +1,28 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:get/get.dart';
+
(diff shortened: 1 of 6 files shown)


jonataslaw:
Wow, this is very nice!
Thank you for that.
In the next version I will increase the tests. I already have some local tests, and this library was really missing. Thank you for your contribution!

chimon2000:
NP.  If you'd like to see a simple way to setup code coverage, check out https://github.com/chimon2000/remote_state/blob/master/.github/workflows/dart.yml

PR #301, opened 2020-06-27
Add extension methods to GetUtils

stefandevo:
I like the idea of the GetUtils, but I also would like them to be usable as an extension method instead through the `GetUtils` class. 
I created extensions methods for all of them.
Hope you like the idea.


jonataslaw:
Great! I was thinking about it right now. You were very precise!
Thanks again.

PR #299, opened 2020-06-27
[Docs] Separating documentation in multiple files, and make main README with less information

Nipodemos:
This is a draft, it should not be merged yet.

Hello @jonataslaw !
I am thinking about this for a few days, but i don't even know if that is a good idea. Since README is getting gigantic and it will be even bigger when everything gets documented, the main readme file is getting bloated with information. It's a lot for one place.

My proposal is to split documentation in different files, make the main README have just a resumed version of all GetX can do, and link for the more extended documentation.

Advantages:
* The first contact of new users with get can be better, improving the amount of people that want to try it out, and making easier for them to do this
* It is easier to maintain, because, if you make an update in just one thing, you can change just one file
* more organization in docs

Disavantages:
* Since it will not have all features on one file, the user can't simply use CTRL + F and search the term he/she wants to learn more
* We will have to focus on two documentations: the resumed version and the extended version

I think it worth it.
If you like the idea, can you give me some advice of what to put in the main README? I'm with little creativity hahah

PR #306, opened 2020-06-28
Create README-es.md

maginkgo:
This is the spanish translation of the Readme.md file, in its latest version (3.1.0)

### DIFF
--- a/README-es.md
+++ b/README-es.md
@@ -1,0 +1,1537 @@
+![](get.png)
+
+*Idiomas: Español (este archivo), [Inglés](README.md), [Portugués de Brasil](README.pt-br.md).*
+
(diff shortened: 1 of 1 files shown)


jonataslaw:
Wow, thank you very much for this PR, it was a great contribution, and I really appreciate it.

Nipodemos:
Hey @maginkgo ,since you also been through the process of messing with the entire documentário, what do you think of the idea on #299 ?
I think the documentation is already too big to be in one file, and it should be separated in multiple files by category, and leave it in the main Readme file only a resumed versions of all the functionalities of GetX 

But I don't have a good idea of how to make a good resumed version to put in main Readme, so I just copied the beginning of each category... But I guess this is not good enough

Do you have some tips or ideas of how to make this better? Or if you think another approach is better, share your thoughts with us :smile:

PR #321, opened 2020-06-30
[Docs] update main README english and pt-br

Nipodemos:
Hello!
From now on whenever i make a new update, i will do one PR per file (putting together when is the same file but different language)

that way it is a lot easier for you to review, and even easier to spot errors or suggest a change

unfortunaly i don't know anything in spanish to be able to update that language


jonataslaw:
LGTM

PR #363, opened 2020-07-09
[docs] more docs for dependency management

Nipodemos:
Hello!

This time i went and documented the options that exist when using Get.put, lazyPut and putAsync

on the main README i've added a few details, which you can see

By the way, there were a few options that i don't know how it works so i can't document it. I've put a TODO in it, but i want to document that before merge

Could explain what those options? so i can update docs

### DIFF
--- a/README.pt-br.md
+++ b/README.pt-br.md
@@ -30,7 +30,7 @@
   - [RxController e GetBuilder se uniram](#rxcontroller-e-getbuilder-se-uniram)
   - [Rotas nomeadas](#rotas-nomeadas)
     - [Porque essa mudança](#porque-essa-mudança)
-- [Porque eu fiz esse package](#porque-eu-fiz-esse-package)
+- [Por que GetX](#por-que-getx)
 
 # Sobre Get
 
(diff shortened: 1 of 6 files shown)


jonataslaw:
Hi Nipodemos, 
"builder" allows you to create the dependency using function instead of the dependency itself. I don't think anyone uses it, and I don't know why I haven't removed it until today.
overrideAbstract allows you after using an abstract class in a test, replace it with another one and follow the test.
I created this, but I also think that it is not something that is so useful in practice.

Fenix is the way that I will activate to avoid 99% of the problems with current dependency injection that people who do not use Bindings, if they create a dependency, and it will be discarded.
Using fenix mode, the dependency is excluded, but if someone needs that controller again, the dependency is created again and delivered.

Nipodemos:
ok, thanks for the answer
i already made all changes i want, can you review please?

README.pt-br.md
```diff
@@ -478,12 +479,11 @@ GetMaterialApp(
 
 # Por que GetX?
 
-1- Muitas vezes após uma atualização do Flutter, muitos dos seus packages irão quebrar. As vezes acontecem erros de compilação, muitas vezes aparecem erros que ainda não existem respostas sobre, e o desenvolvedor necessita saber de onde o erro veio, rastrear o erro, para só então tentar abrir uma issue no repositório correspondente, e ver seu problema resolvido. Get centraliza os principais recursos para o desenvolvimento (Gerencia de estado, de dependencias e de rotas), permitindo você adicionar um único package em seu pubspec, e começar a trabalhar. Após uma atualização do Flutter, a única coisa que você precisa fazer é atualizar a dependencia do Get, e começar a trabalhar. Get também resolve problemas de compatibilidade. Quantas vezes uma versão de um package não é compatível com a versão de outro, porque um utiliza uma dependencia em uma versão, e o outro em outra versão? Essa também não é uma preocupação usando Get, já que tudo está no mesmo package e é totalmente compatível.
```
Nipodemos, line 481: essa parte do "porque getx" eu não mexi em nada, não sei porque eles tão considerando que tem um diff

Nipodemos:
@jonataslaw meu PR já está pronto pra review, você poderia dar uma olhada nele por favor?

README.pt-br.md
jonataslaw: Github as vezes dá a louca

PR #371, opened 2020-07-13
Update link for state management explanation

danielolaviobr:
Fixing the link for the detailed explanation on the state management section


jonataslaw:
Hi @danielolaviobr, Thanks for your contributation!
LGTM

PR #374, opened 2020-07-13
Add util extensions again in the lib + add missing function

stefandevo:
The extensions were removed from the export


stefandevo:
FYI Build was not run due to Github outtage

jonataslaw:
Thank you for another contribution.
I tested it locally and everything is ok.
LGTM

PR #375, opened 2020-07-13
[docs] Documenting main methods

Nipodemos:
Hello again!

yesterday i decided that would be a good idea to have the methods that we use a lot documented, because is nice when we hover our mouse on method and a good description show up.

However, i noticed that many things i don't know what is for, so i need to ask you for help 😅 

- What is that `id` in basically everything related to routes? Could you explain it's use?
- what exactly `fullScreenDialog` do? i mean, i can imagine a little with the name, but i don't know for sure what's that
- what about popGesture? i am imagining that is for enable pop a page by drag, but i want to make sure too
- a read documentation about those predicate things, but i wondering if you can use it normally with get. They were showing that you should use `ModalRoute`, but how to adapt this to Get? or it doesn't need adaptation?
- what is the difference of `offAndToNamed()` and `offNamed()`? 
- `offAndToNamed()` has a result argument. What is it for?

i guess that's all for now

Off-topic: Where all default configs of get are being defined? like default transition and default trasition duration? I'm just curious as i couldn't find at first glance in your code


marcosfons:
I think this id in route methods is related to NestedNavigators. Therefore, when you create a NestedNavigator, you create a new key with Get.nestedKey(keyId) and, when push a new route, pass that keyId as a parameter for Get know which key you want use

Nipodemos:
oh that's true, i hadn't noticed that documentation, thank you @marcosfons  for pointing that out.

pretty crazy that nested navigation thing

now @jonataslaw , when you have free time could you help me with the other infos?

jonataslaw:
> Hello again!
> 
> yesterday i decided that would be a good idea to have the methods that we use a lot documented, because is nice when we hover our mouse on method and a good description show up.
> 
> However, i noticed that many things i don't know what is for, so i need to ask you for help 
> 
> * What is that `id` in basically everything related to routes? Could you explain it's use?
the id is used in nested navigation, as pointed by marcosfons 

> * what exactly `fullScreenDialog` do? i mean, i can imagine a little with the name, but i don't know for sure what's that

Flutter allows you to open routes "fullScreenDialog", which looks more like a dialog with an "x" in the upper corner to close, it is a navigation feature of Flutter itself, it is not something of Getx.

> * what about popGesture? i am imagining that is for enable pop a page by drag, but i want to make sure too

On iOS when you drag a page from left to right you close it.
I had to create this feature in the routes, so I ended up opening the possibility to use it on android too.

> * a read documentation about those predicate things, but i wondering if you can use it normally with get. They were showing that you should use `ModalRoute`, but how to adapt this to Get? or it doesn't need adaptation?

You can use ModalRoute as not.
You can use a "Get.currentRoute == '/home'"
You can create the predicate as you wish, and not just with modalRoute.

> * what is the difference of `offAndToNamed()` and `offNamed()`?

The first one page back, and goes to the next. The second goes to the next one, and removes the previous one. The route transition animation is different.

> * `offAndToNamed()` has a result argument. What is it for?

It is the same as using a Get.back() and then a Get.to().
In the same way that you could return an argument to the previous route, and go to the next one, this is also possible.

> i guess that's all for now
> 
> Off-topic: Where all default configs of get are being defined? like default transition and default trasition duration? I'm just curious as i couldn't find at first glance in your code

Here:
https://github.com/jonataslaw/getx/blob/master/lib/src/get_main.dart#L26

lib/src/get_main.dart
```diff
@@ -29,54 +29,76 @@ class GetImpl implements GetService {
...
+  /// Pushes a new [page] to the stack
+  ///
+  /// It has the advantage of not needing context, so you can call from your business logic
+  ///
+  /// You can set a custom [transition], a transition [duration].
+  /// You can send any type of value to the other route in the [arguments].
+  /// If you're using the [Bindings] api, it's here that you use it
+  ///
+  /// It replaces Navigator.push and it doesn't have the Navigator.push
   /// routes rebuild bug present in Flutter. If for some strange reason you want the default behavior
```
jonataslaw, line 41: This bug has not existed since version 1.17.
I think this can now be removed. I don't know if I can change this in your branch, would you be able to remove this please for me?

stefandevo:
Sorry to break in this PR; @jonataslaw could you have a look at the Slack channel regularly for some questions from members? I know and understand you are very busy. I also asked something in private message for github integrations. Sorry to disturb.

Nipodemos:
@jonataslaw i've finished it.

Added everything you said.

**Just a note: i haven't changed one letter of code, i've just added commas to help in code organization**

To me it is ready to merge, just need your approval

jonataslaw:
Thank you, it looks very good.
I'll merge.

Issue #378, opened 2020-07-14
Documentation | Link not proper


NiranjanShah:
Link to Route management goes to Dependency Management.

Nipodemos:
I've looked for this error, but i could not find it. In none of the three languages. Could you point specificly where is the typo?

NiranjanShah:
its not a typo, Under topics -> The Three pillars -> Route Management -> In depth Explanation. THIS takes to Dependency Management i/o Route Management.

Nipodemos:
![lg9oLvsh1I](https://user-images.githubusercontent.com/11494727/87460367-278cac00-c5e3-11ea-826c-851bda293287.gif)

i just made that gif, it does lead to the correct page

NiranjanShah:
Oh, sorry, as the line you are clicking is in the end of the Route Management paragraph, the confusion raised.

After in depth explanation and next visible topic is Dependency Management.

Nipodemos:
no problem man, i understand. I'll see if anything can be done to be more intuitive, docs still has a lot to improve

Nipodemos:
since it is solved, closing

PR #379, opened 2020-07-14
[docs] several changes in docs

Nipodemos:
Hello again!
I noticed a few things that could change:

### The State manager part was not truly explaing state management with get.
Instead it had only the example that covers the 3 pillars of GetX. 

So i decided to create a new topic just for the counter app, and put a new resumed part about state manager.

In order to keep the smallest possible, i've put only reactive state manager. This is also to incentivate people to use more reactive state manager instead of simple while still leaving them with full choice

### I forgot to put videos in documentation, so here they are

Amateur coder did two great videos: one about route management, and another about state management. I've put each one in their separate topics.

@tadaspetra i did not include your video on the first impressions about GetX because the two you made recently fit the most in documentation, i hope you don't mind

additionally, @jonataslaw i've put a link to your corona app video in the PT_BR docs, since the video is in portuguese. Unfortunally it is not so updated anymore, i think in the video you used `update(this)`, so i wonder if it is a good idea to keep the link or not. I think it is still a good idea

### There were some little parts in english readme that weren't in other languages
so i've added

### Now the only place where the link to in-depth explanation appear is as the end of topic
after #378 , i think is a better idea not to have duplicated links like that. This way i guess it got better

### updated all TOC and add new ones

i've noticed that the separeted docs file didn't had their own TOC, and it is always good to have one, especially in State Management section

that's all

@jonataslaw if you have any change, just tell me and i'll do 😄


tadaspetra:
This is awesome! And that was my intent. The first video was just me exploring the topic. These next two were the ones where I looked deeper into it to understand it better. I have one more coming up on friday that I'm working on to complete the "trilogy" called about "Other GetX Features" which will include binding, localization, and a bunch of other things :)

jonataslaw:
That video I made is very outdated, very much!
I think it's better not to include it, even I think about disabling that video.

Nipodemos:
ok @jonataslaw , i removed the video link as suggested
ready to merge, just need your approval

jonataslaw:
Thanks for this!

PR #386, opened 2020-07-15
add function for cnpj validation

kauemurakami:
Adding function for cnpj validation


stefandevo:
For each function you need to add the appropriate extension function also.

kauemurakami:
@stefandevo 
Is already done !! Thanks for the guidance!

lib/src/regex/get_utils_extensions.dart
```diff
@@ -32,6 +32,7 @@ extension GetStringUtils on String {
   bool get isPassport => GetUtils.isPassport(this);
   bool get isCurrency => GetUtils.isCurrency(this);
   bool isCpf(String s) => GetUtils.isCpf(this);
+  bool isCnpj(String s) => GetUtils.isCpf(this);
```
jonataslaw: bool isCnpj(String s) => GetUtils.isCnpj(this); 

please, fix the second parameter

PR #394, opened 2020-07-17
Translation return null value when no entry found

jasonlaw:
There are 2 changes in this PR:
1. A new method appendTranslations. 
   Reason: I have a shared package with all the common widgets for several projects. In this package I will add the translations into Get. However, the translations will be replaced by the project with addTranslations.

 2. Fixed to prevent return null value if there is no translation found, by checking the key existence explicitly. Also remove the "use the first language code values" if no such language code is added into translations. I think simply return the original value would be more making sense. Hope this is fine for you.

### DIFF
--- a/lib/src/get_main.dart
+++ b/lib/src/get_main.dart
@@ -655,6 +655,16 @@ class GetImpl implements GetService {
     translations.addAll(tr);
   }
 
+  void appendTranslations(Map<String, Map<String, String>> tr) {
+    tr.forEach((key, map) {
+      if (Get.translations.containsKey(key)) {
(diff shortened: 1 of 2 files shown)


Nipodemos:
i think if possible, to make return the first language code value, and if non-existent, return the key itself as you put.

I think is better to display the string in the main language instead of leave the key there, which have no meaning to the user

jasonlaw:
The key is actually the main language itself, in my case is English. If say we have other language Chinese and Melayu maintain, and say user switch to Melayu but not entry found, it will display Chinese, this is weird, it should be return the key which is maintained in English for my case. Does this make sense?

PR #397, opened 2020-07-17
Updated docs for the last GetX video

tadaspetra:
I though it would be good to add the last of the 3 part series going over the GetX basics.

I couldn't figure out how to update the Table of Contents lol. I started googling and it seems like this might be using something that generates one for you? https://github.com/ekalinin/github-markdown-toc

If I was supposed to do something different please let me know!

This is my first time contributing to a project that isn't my own. So I'm not too good with GitHub yet 😂. Please forgive me!

### DIFF
--- a/README-es.md
+++ b/README-es.md
@@ -22,6 +22,7 @@
   - [Gestión del Estado](#gestión-del-estado)
     - [Reactivo STATE_MANAGER](#reactivo-state_manager)
     - [Más detalles sobre la gestión del estado.](#más-detalles-sobre-la-gestión-del-estado)
+    - [Video explanation about state management](#video-explanation-about-state-management)
   - [Gestión de Rutas](#gestión-de-rutas)
     - [Más detalles sobre la gestión de rutas.](#más-detalles-sobre-la-gestión-de-rutas)
     - [Video Explanation](#video-explanation)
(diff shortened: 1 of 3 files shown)


Nipodemos:
hello! thank you for this 😄 

In order to update the TOC i've used an extension from VS code, [Markdown all in one](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one)

it have a command to create automatically the TOC
but @jonataslaw want to have the main topics of TOC to be bold, and the extension doesn't allow that customization, so you have to make bold manually after generating, which is not hard at all

Please, it would be nice if you post that change in docs on the spanish and portuguese translations, you can remain them in english. In my opinion, is best to have the description untranslated then don't have (imagine you make the documentation of some feature and that docs exist only in english, who reads the portuguese one wouldn't even have a clue of that documentation existence. If it is there in english at least they know it exists and can use google translate or something)

since you said you are still new to github, if you find hard to update your pull request, i can do it later.

Thanks for this contribution!

tadaspetra:
No problem. I will update this tomorrow. Thank you for the feedback!

jonataslaw:
Thanks for this. 
LGTM, I'll merge

PR #409, opened 2020-07-20
Get Create finalized

PauloPortfolio:
# Description
Get Create (Get.create) was almost done; however, it was not able to return the instance created.

## Type of change:
- [ ] Get Create feature was finalized/completed  (non-breaking changes)
- [ ] This change requires a documentation update

## Files changed:
- [ ] get_instance.dart
- [ ] extension_instance.dart
- [ ] get_instance_test.dart

## Tests created:
- [ ] Get.create test

### DIFF
--- a/lib/src/instance/extension_instance.dart
+++ b/lib/src/instance/extension_instance.dart
@@ -11,7 +11,7 @@ extension Inst on GetImpl {
           {String tag, bool permanent = false}) async =>
       GetInstance().putAsync<S>(builder, tag: tag, permanent: permanent);
 
-  void create<S>(FcBuilderFunc<S> builder,
+  S create<S>(FcBuilderFunc<S> builder,
           {String name, bool permanent = true}) =>
       GetInstance().create<S>(builder, name: name, permanent: permanent);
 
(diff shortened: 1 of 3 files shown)


jonataslaw:
Wow, thanks for your contribution.
I just need to add a test to see if this will be the expected behavior, and then I will merge.

PauloPortfolio:
Let me know about the test you have mentioned. Unfortunately, currently, I am not proficient in tests in Dart; however, overtime, I will figure out that.

Nipodemos:
What that create method do?

PauloPortfolio:
@Nipodemos it simply create a non-singleton instance.

jonataslaw:
Hi Paulo, first I would like to thank you for the time dedicated to this contribution, but I just wanted to understand the reason for inserting a find inside create, could you give me a situation of direct use of it?
I also recently modified GetWidget to have a cache, which means that if you do something like:

Get.create<Controller>(() => Controller());

in your Binding, you can simply do:

class Home extends GetWidget<Controller> {
}

And an instance of create will be created and made available only for this screen, this can be useful in your projects.

joanofdart:
I think the situation is just that upon creating an instance, it returns it? 🤔

PauloPortfolio:
Create a simple  dart instance individually would be the solution.

PR #415, opened 2020-07-22
formatting files according to dartfmt

MSimone12:
# Formatting files listed as bad formatted in pub.dev

## Files changed:

* lib/src/regex/get_utils.dart
* lib/src/root/root_widget.dart

@jonataslaw no issue open related to this PR, but only this is not allowing to get package in pub.dev reaching max points


jonataslaw:
Thanks!

Issue #417, opened 2020-07-23
Navigator popUntill() Equivalent needed. 


manojeeva:
**Is your feature request related to a problem? Please describe.**
Navigator popUntill method is used to just pop routes from the stack.
In Get offNamedUntill or offNamed or back or offAllNamed are removing some routes and pushing a new route to stack.
What if we need to only pop a few routes from the stack?

**Describe the solution you'd like**
We can have the same as the Navigator popUntill method to remove some routes.

**Describe alternatives you've considered**
 ```Navigator.of(context).popUntil((route) => route.settings.name == '/two');```
current flow
``` Get.offNamedUntil("/two", (route) => route.settings.name=='/two'); ```
changed to 
```     Get.offNamedUntil((route) => route.settings.name=='/two',optional(newroute)); ```
offNamedUntill method can be used. the predicate should be the first argument and 'newrouteName' should be optional If 'newrouteName'  is not supplied we can simply pop the routes using predicate

jonataslaw:
Get.until();

manojeeva:
> Get.until();
Thanks.
 Is it mentioned in doc or readme file? It will be nicer to see all the functionalities.
Can you refer to something to check all the functionality?

Nipodemos:
@manojeeva yes is is 😄 
![image](https://user-images.githubusercontent.com/11494727/88300328-baab9d00-ccd9-11ea-855e-6ec03b0ea86a.png)

you can also type the method in vscode and read the docs about the method
![image](https://user-images.githubusercontent.com/11494727/88300791-41f91080-ccda-11ea-863b-36e6b948b34c.png)

i just noticed that there is an typo in docs,😅

PR #421, opened 2020-07-23
fix typo of predicate arguments in Get route methods

Nipodemos:
Just noticed in #417 that i documented wrongly the predicate argument in all methods that use it. Fixing.

### DIFF
--- a/lib/src/get_main.dart
+++ b/lib/src/get_main.dart
@@ -144,7 +144,7 @@ class GetImpl implements GetService {
   /// as explained in documentation
   ///
   /// [predicate] can be used like this:
-  /// `Get.until(Get.currentRoute == '/home')`so when you get to home page,
+  /// `Get.until((route) => Get.currentRoute == '/home')`so when you get to home page,
   ///
   /// or also like this:
   /// `Get.until(!Get.isDialogOpen())`, to make sure the dialog is closed
@@ -154,17 +154,19 @@ class GetImpl implements GetService {
     return global(id).currentState.popUntil(predicate);
   }
 
-  /// Push the given [page], and then pop several [pages] in the stack until
+  /// Push the given [page], and then pop several pages in the stack until
   /// [predicate] returns true
   ///
   /// [id] is for when you are using nested navigation,
   /// as explained in documentation
   ///
+  /// TODO: Learn how to get a Route class to use with this method
+  ///
   /// [predicate] can be used like this:
-  /// `Get.until(Get.currentRoute == '/home')`so when you get to home page,
+  /// `Get.until((route) => Get.currentRoute == '/home')` so when you get to home page,
   ///
   /// or also like this:
-  /// `Get.until(!Get.isDialogOpen())`, to make sure the dialog is closed
+  /// `Get.Until((route) => !Get.isDialogOpen())` to make sure the dialog is closed
   Future<T> offUntil<T>(Route<T> page, RoutePredicate predicate, {int id}) {
     // if (key.currentState.mounted) // add this if appear problems on future with route navigate
     // when widget don't mounted
@@ -180,9 +182,9 @@ class GetImpl implements GetService {
   /// as explained in documentation
   ///
   /// [predicate] can be used like this:
-  /// `Get.until(Get.currentRoute == '/home')`so when you get to home page,
+  /// `Get.until((route) => Get.currentRoute == '/home')`so when you get to home page,
   /// or also like
-  /// `Get.until(!Get.isDialogOpen())`, to make sure the dialog is closed
+  /// `Get.until((route) => !Get.isDialogOpen())`, to make sure the dialog is closed
   ///
   /// Note: Always put a slash on the route ('/page1'), to avoid unnexpected errors
   Future<T> offNamedUntil<T>(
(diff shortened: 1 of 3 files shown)


lib/src/get_main.dart
```diff
@@ -154,17 +154,19 @@ class GetImpl implements GetService {
...
   }
 
-  /// Push the given [page], and then pop several [pages] in the stack until
+  /// Push the given [page], and then pop several pages in the stack until
   /// [predicate] returns true
   ///
   /// [id] is for when you are using nested navigation,
   /// as explained in documentation
   ///
+  /// TODO: Learn how to get a Route class to use with this method
```
Nipodemos, line 163: Also, i've put a todo in docs, i don't know how would i get a Route method in Get, can you explain that to me?

lib/src/get_main.dart
```diff
@@ -144,7 +144,7 @@ class GetImpl implements GetService {
   /// as explained in documentation
   ///
   /// [predicate] can be used like this:
-  /// `Get.until(Get.currentRoute == '/home')`so when you get to home page,
+  /// `Get.until((route) => Get.currentRoute == '/home')`so when you get to home page,
```
jonataslaw, line 147: Nice!

lib/src/get_main.dart
jonataslaw: In fact, he doesn't need a route class, just a VoidCallback where he delivers the route and gets a response bool.

(route)=> for example, route is given by callback, so you can do something like:

Get.until((route) => route.isFirst)

Nipodemos: but that is not the predicate argument?

`Get.until()` receives as first argument predicate, which you told me once is exactaly that, a VoidCallback

but `Get.offUntil()` has a `Route<T> page` and also the `RoutePredicate predicate`
i have doubts abou the first one

Nipodemos:
hey i think you have forgot me kkkkkk

i still have a doubt that i replied on that piece of code, let me replicate here:

> but that is not the predicate argument?
> 
> `Get.until()` receives as first argument predicate, which you told me once is exactaly that, a VoidCallback
> 
> but `Get.offUntil()` has a `Route<T> page` and also the `RoutePredicate predicate`
i have doubts abou the first one

jonataslaw:
> hey i think you have forgot me kkkkkk
> 
> i still have a doubt that i replied on that piece of code, let me replicate here:
> 
> > but that is not the predicate argument?
> > `Get.until()` receives as first argument predicate, which you told me once is exactaly that, a VoidCallback
> > but `Get.offUntil()` has a `Route<T> page` and also the `RoutePredicate predicate`
> > i have doubts abou the first one

Predicate is a function that performs a condition when false, and does not execute it when true. It has a VoidCallback that will receive a Route<T>.

You can for example use (_) => false as a predicate so that this function is executed whenever the loop is started. Get.offAll for example, used this before version 1.17.

Get.offUntil() will close the page, and proceed to the next Route<T>.
It will do this shortly after closing all pages according to the predicate. Imagine that you have a testing app. You entered the home, clicked on math, and started the exam. You want that when the user finishes the last question, he moves to the score / note page, but you want all pages of the exam to be removed from the stack (because it makes no sense to allow the user to go back to the questions if he has already finished the exam).
Then, you will go to the result page (Route) and close all pages until home ((route) => Get.currentRoute == '/ home').
If you did not understand the operation, I will explain in Portuguese, it may be easier:


///////////////////// PORTUGUESE VERSION:

Predicado é uma função que executa uma condição quando falsa, e não a executa quando verdadeira. Ela tem um VoidCallback que irrá receber uma Route<T>. 

Você pode por exemplo usar (_)=> false como predicado para que essa função seja executada sempre que o loop for iniciado. Get.offAll por exemplo, usava isso antes da versão 1.17. 

Get.offUntil() irá fechar a página, e avançar para a próxima Route<T>. 
Ela irá fazer isso, logo após, fechar todas as páginas de acordo com o predicado. Imagine que você tem um app de testes. Você entrou na home, clicou em matemática, e iniciou o exame. Você quer que quando o usuário termine a última questão, ele avance para a página de pontuação/nota, mas você quer que todas páginas do exame sejam removidas da stack (porque não faz sentido permitir o usuário que volte para as questões se ele já terminou o exame).
Logo, você irá ir para página do resultado (Route) e irá fechar todas as páginas até a home ( (route)=> Get.currentRoute == '/home' ).

Nipodemos:
what i want to make sure is the `page` argument, because it is not of type Widget, how should a pass a widget to `Get.offUntil()`? @jonataslaw

jonataslaw:
> what i want to make sure is the `page` argument, because it is not of type Widget, how should a pass a widget to `Get.offUntil()`? @jonataslaw

GetPageRoute(page:()=> NextPage())

jonataslaw:
Maybe it would be easier to copy the code from Get.to, and send a Widget and use parameters, but that would be a break update, so it would not be possible to do this in version 3.

Nipodemos:
> Maybe it would be easier to copy the code from Get.to, and send a Widget and use parameters, but that would be a break update, so it would not be possible to do this in version 3.

makes sense, i'm fine with that.



> > what i want to make sure is the `page` argument, because it is not of type Widget, how should a pass a widget to `Get.offUntil()`? @jonataslaw
> 
> GetPageRoute(page:()=> NextPage())

Thank you very much for the explanation, it was exactly what i wanted. I'll update the code and fix the conflicts

Nipodemos:
oh gosh everything is in a differente place, i'll have to make another PR

Issue #430, opened 2020-07-29
Default translation fallback


nemoryoliver:
**Is your feature request related to a problem? Please describe.**
Not related to a problem

**Describe the solution you'd like**
Hi Jonny,

Once again, thanks for your super awesome work on **Get**! I have been using it for my apps now and works awesome so far. My only wished feature right now is the ability to set a fallback translation.

Example I have these:

```
en_US = {'hi': 'Hi', 'hello': 'Hello there'}
es_ES = {'hi': 'Hola'}
```

 but **es_ES** does not have the 'hello' key. so if I use `'hello'.trArgs()` I wish that i will be able to return the fallback from the default **en_US** which is equal to **'Hello there'**. Right now the default behavior is that it will return the **key** `String` which is **'hello'**

jonataslaw:
Added on https://github.com/jonataslaw/getx/pull/438
Thanks!
I'm closing this.

nemoryoliver:
Thank you very much!

PR #438, opened 2020-08-01
fallbackLocale for translation (as GetMaterialApp config)

lundin:
Hi, what do you think about this as a solution to issue at [https://github.com/jonataslaw/getx/issues/430](url)
Doing this will allow a specific fallbackLocale. I cant do any tests but tested with the something key translation and it showed "still something" when it failed to find any in nl,BE or nl.

```
 locale: Locale('nl', 'BE'),
 fallbackLocale: Locale('en', 'US'),
 translationsKeys: {
        'nl_BE': {'title': 'Hallo van GetX'},
        'en_US': {'title': 'Hello from GetX', "something": "still something"}
      },
```

### DIFF
--- a/lib/src/get_interface.dart
+++ b/lib/src/get_interface.dart
@@ -33,6 +33,8 @@ abstract class GetInterface {
   GetMaterialController getxController = GetMaterialController();
 
   Locale locale;
+  
+  Locale fallbackLocale;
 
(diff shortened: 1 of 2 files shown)


jonataslaw:
Hi, thanks for that contribution!
It looks very good.
I'll review this and then merge

nemoryoliver:
Thank you!!!

PR #453, opened 2020-08-07
test: add controller's onInit test

tiagocpeixoto:
Add tests to check if the controller's onInit is being invoked by Get.put after Get.delete.


Nipodemos:
nice!

jonataslaw:
Thank you!
I appreciated that a lot.

PR #481, opened 2020-08-19
Abstraction of logging

stefandevo:
Hi

There is some logging in the package currently, all simple `print` statements.

However, when integrating the getX package with logging libraries it would be nice to be able to capture these log lines in once place.

Therefore I created an abstraction of the log print statements, into a Callback. Everywhere where you used the `print` statement and the check if the log is enabled, I replaced it with a single statement like `GetConfig.log('[GETX] $key has been initialized');`

A few locations had a `print` statement for errors too (no check if logging is enabled). This is replaced with the same but with the optional parameter `GetConfig.log('Instance $newKey not found', isError: true);`.

By doing this, it is now possible to replace the logging by a log provider the dev chooses.

I changed (and tested) it in the sample application.

```
GetMaterialApp(
      debugShowCheckedModeBanner: false,
      initialRoute: '/',
      enableLog: true,
      logWriterCallback: localLogWriter,
      getPages: [
        GetPage(name: '/', page: () => HomePage(), binding: HomeBinding()),
        GetPage(name: '/country', page: () => CountryPage()),
        GetPage(name: '/details', page: () => DetailsPage()),
      ],
    ),
```

and define your log writer:

```
void localLogWriter(String text, {bool isError = false}) {
  // use your own log library here -- for demo, just the same print
  print('** ' + text + ' [' + isError.toString() + ']');
}
```

Of course this is optional. You can remove the `logWriterCallback` assignment to use the default implementation.

I hope you like the change. This would be really helpful for myself to route the logs to my own log provider.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -10,6 +10,8 @@ void main() {
     GetMaterialApp(
       debugShowCheckedModeBanner: false,
       initialRoute: '/',
+      enableLog: true,
+      // logWriterCallback: localLogWriter,
       getPages: [
         GetPage(name: '/', page: () => HomePage(), binding: HomeBinding()),
         GetPage(name: '/country', page: () => CountryPage()),
@@ -17,4 +19,9 @@ void main() {
       ],
     ),
   );
+}
+
+// Sample of abstract logging function
+void localLogWriter(String text, {bool isError = false}) {
+  print('** ' + text + ' [' + isError.toString() + ']');
 }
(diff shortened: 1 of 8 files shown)


jonataslaw:
That looks so cool!
I was thinking about it, but unfortunately I didn't have time to implement it. Thank you very much Stefan!

ghenry:
What logging library do folks recommend here or a good list to review? Thanks.

PR #484, opened 2020-08-20
Docs updated for logWriterCallback option

stefandevo:
Added docs for the logging option

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,22 @@ Get.config(
 )
 ```
 
+You can optionally redirect all the logging messages from Get. If you want to use your own favourite logging package and want to capture the logs there.
+
+```dart
+GetMaterialApp(
+  enableLog: true,
+  logWriterCallback: localLogWriter,
+);
+
+void localLogWriter(String text, {bool isError = false}) {
+  // pass the message to your favourite logging package here
+  // please note that even if enableLog: false log messages will be pushed in this callback
+  // you get check the flag if you want through GetConfig.isLogEnable
+}
+
+```
+
 ## Video explanation of Other GetX Features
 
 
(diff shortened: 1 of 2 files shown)


stefandevo:
@Nipodemos maybe you can update the -BR version too?

stefandevo:
@ajatsoft maybe you can update the -ES version too?

stefandevo:
@szymon-filipiak maybe you can update the -PL version too?

Nipodemos:
i noticed that one of the best ways to truly understand a topic is to translate it. 

only after i translated that i've noticed  that this logWriter does not respect `Get.isLogEnable`. Why?

stefandevo:
Because then you can control it from the logging package to decide to collect the log. 
Also then there is also one location in the library to check if logging is needed (see log.dart).

Nipodemos:
oh i understand, thanks for the reply.
@jonataslaw since the other guys will need to make a new PR to translate, i think this can be approved

now i am wondering if we should put this docs in the other languages files the english version so that someone that will read the translated docs can at least acknowledge that there is a logWriterCallback.

What do you think @stefandevo , i'm wondering this for ages

ajatsoft:
Yes, I will try to do this night

jonataslaw:
> oh i understand, thanks for the reply.
> @jonataslaw since the other guys will need to make a new PR to translate, i think this can be approved
> 
> now i am wondering if we should put this docs in the other languages files the english version so that someone that will read the translated docs can at least acknowledge that there is a logWriterCallback.
> 
> What do you think @stefandevo , i'm wondering this for ages

I agree with the approval of this PR, as well as the inclusion, although English in the other docs

jonataslaw:
I'll merge, then we insert the English version in the other docs while the Spanish and Polish versions are not ready.

ajatsoft:
I update the Spanish version with it in last pr

PR #501, opened 2020-08-25
Added 2 new widgets to manage local state and some cleanup

roipeker:
- commented Obx()
- close subscription in Obx
- added the experimental ObxValue() widget to manage local states based on Rx.
- added experimental ValueBuilder(), analogue to ObxValue(), but uses a simpler StatefulWidget with a callback.

### DIFF
--- a/lib/src/state_manager/rx/rx_obx.dart
+++ b/lib/src/state_manager/rx/rx_obx.dart
@@ -1,5 +1,8 @@
+import 'dart:async';
+
 import 'package:flutter/widgets.dart';
 import 'package:get/src/state_manager/rx/rx_interface.dart';
+
 import 'rx_impl.dart';
 
 Widget obx(Widget Function() builder) {
@@ -24,6 +27,13 @@ class Obxx extends StatelessWidget {
   }
 }
 
+/// The simplest reactive widget in GetX.
+///
+/// Just pass your Rx variable in the root scope of the callback to have it
+/// automatically registered for changes.
+///
+/// final _name = "GetX".obs;
+/// Obx(() => Text( _name.value )),... ;
 class Obx extends StatefulWidget {
   final Widget Function() builder;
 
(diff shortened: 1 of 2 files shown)


jonataslaw:
Thank you for that contribution.
I can't wait to test this out.

Nipodemos:
Hello!
Could you explain me the with more details what those two widgets are for? i read the comments in the code but i didn't understand very well 😅 
What are their use cases?

lib/src/state_manager/rx/rx_obx.dart
```diff
@@ -71,3 +83,59 @@ class _ObxState extends State<Obx> {
   @override
   Widget build(BuildContext context) => notifyChilds;
 }
+
+/// Similar to Obx, but manages a local state.
+/// Pass the initial data in constructor.
+/// Useful for simple local states, like toggles, visibility, themes,
+/// button states, etc.
+///  Sample:
+///    ObxValue((data) => Switch(
```
Nipodemos, line 92: For what i understood, this example should have the casting `ObxValue<Controller>()`, is this correct?

jonataslaw:
> Hello!
> Could you explain me the with more details what those two widgets are for? i read the comments in the code but i didn't understand very well 
> What are their use cases?

This is for ephemeral state control, when you don't want and don't need to create a controller, you just want to auto update a widget, as in the example of a Switch.

lib/src/state_manager/rx/rx_obx.dart
roipeker: The generic type the part that fails when defined, my bad, i think i left a TODO for it. Im waiting @jonataslaw feedback, to avoid the exception... and I will make other samples soon.
This is not relates to a controller, uses a local state with a Rx value directly. For instance, to toggle a button for toggleVisibility in a textfield, changing a Theme, use an integer for tab navigation/bottom navigation... those type of simple local variables.

odunboye: Can this work with Something like VideoPlayerController, Say you would like to initialise a VideoPlayerController for Playback?

PR #516, opened 2020-08-27
improvement of logging, replaced print method for developer.log

jorgegaticav:


stefandevo:
- Is this backwards compatible?
- Inconsistent name used; sometime getx then GETX
- You need to update the docs too in the readme

stefandevo:
Looking further in detail. 
I don't think the signature of the abstract class shld be changed and extended with the name parameter. 

I do agree with removing the [GETX] from the logs. Then the implementation of the default logger can be changed to the developer method adding GETX as the name inside the method. 

This does not break anything and simplyfies it. There is no need to pass the name GETX in the whole library and override implementation of the log callback can add the GETX again in the message if needed.

jorgegaticav:
Hi @stefandevo, I agree with your suggestions. I made the changes. Thank you for your feedback.

jonataslaw:
I reviewed this, and LGTM now.
I'll merge.

jonataslaw:
Thanks for your contributation @jorgegaticav !

PR #519, opened 2020-08-27
GetInstance cleanup

roipeker:
- Cleanup unused _instance_ parameter in `GetInstance::find()`
- Cleanup log messages to have a more consistent format (still work to be done there).


lib/src/instance/extension_instance.dart
```diff
@@ -10,13 +11,36 @@ extension Inst on GetInterface {
...
+  /// Creates a new Instance<[S]> from the <[S]>[builder] callback.
+  /// Every time [find]<[S]>() is used, it calls the builder method to generate
+  /// a new Instance [S].
+  ///
+  /// Example:
+  ///
+  /// ```create(() => Repl());
+  /// Repl a = find();
+  /// Repl b = find();
+  /// print(a==b); (false)```
```
Nipodemos, line 23:
```suggestion
  /// ```
  /// create(() => Repl());
  /// Repl a = find();
  /// Repl b = find();
  /// print(a==b); (false)
  ///```
```

just a small suggestion, i think is better like that
the rest of the code i don't have the knowledge to review 😅

lib/src/instance/get_instance.dart
```diff
@@ -54,12 +62,17 @@ class GetInstance {
...
-  /// Create a new instance from builder class
-  /// Example
-  /// create(() => Repl());
+  /// Creates a new Class Instance [S] from the builder callback[S].
+  /// Every time [find]<[S]>() is used, it calls the builder method to generate
+  /// a new Instance [S].
+  ///
+  /// Example:
+  ///
+  /// ```create(() => Repl());
```
Nipodemos, line 71:
```suggestion
  /// ```
  /// create(() => Repl());
```

lib/src/instance/get_instance.dart
```diff
@@ -54,12 +62,17 @@ class GetInstance {
...
   /// Repl a = find();
   /// Repl b = find();
-  /// print(a==b); (false)
+  /// print(a==b); (false)```
```
Nipodemos, line 74:
```suggestion
  /// print(a==b); (false)
  ///```
```

lib/src/instance/get_instance.dart
```diff
@@ -69,21 +82,26 @@ class GetInstance {
...
-
+    final key = _getKey(S, name);
     _singl.putIfAbsent(
         key, () => FcBuilder<S>(isSingleton, builder, permanent, false));
   }
 
+  /// Clears from memory registered Instances associated with [routeName] when
+  /// using [GetConfig.smartManagement] as [SmartManagement.full] or [SmartManagement.keepFactory]
+  /// Meant for internal usage of [GetPageRoute] and [GetDialogRoute]
+  ///
```
Nipodemos, line 102: This is just a aesthetical suggestion, but i think is not nice to leave a blank line between the docs and the methods that is being documented.
Since the doc are directly connected to the method, they should be close. 
But that's just my opinion, if you prefer the way it is (and id fine for @jonataslaw ), it can remain as is.

stefandevo:
Pls note that there is already a pr pending that will change all log messages. Dif merge needs to be done.

roipeker:
> Pls note that there is already a pr pending that will change all log messages. Dif merge needs to be done.

So, merging will be complicated? I am not used to do PRs.

lib/src/instance/get_instance.dart
roipeker: No problem, I didn't see any standards for it, code was more readable for me. 
It would cool to have some feedback about it though, if it can work as a foundation for a proper docs later at least.

Nipodemos:
i can help with that fix conflicts part, i don't mind

jonataslaw:
I just tested it, and it's a big change to code security.
Now it is no longer possible to override onStart, which is something I always thought to prevent, as it directly affects the call of onInit and onReady.
Many methods that should not be accessed have become inaccessible, as they influence SmartManagement.
This was exactly the code change I was planning to make, so I need to thank you for that, because it makes the code so much more reliable.
When it comes to documentation, I just liked each piece of writing, I just agree to remove the last "///" as suggested by @nipodemos.
As for the other pending PR, @Stefandevo is right.
GetConfig.log ('[GETX] "$ newKey" deleted from memory');
Now it will be:
GetConfig.log ("$ newKey" deleted from memory ');

As this PR depends on https://github.com/jonataslaw/getx/pull/516
I will accept the PR that normalizes the logs, and then all "[GETX]" can be removed from the code (along with the blank lines) and that PR will be ready to merge.

Nipodemos:
i've fixed the conflicts, to my surprise this one was easy 😅 
just missing the blank lines removal part and the suggestions

roipeker:
@Nipodemos The only change in _get_instance.dart_ were the [GetX] tags, right?

roipeker:
@Nipodemos, @jonataslaw would you mind trying the changes included in the last commits?
all Rx values have a  >>=  and a toString() that resolves to the rx.value.
Also give a try to the new BindingsBuilder((){}) class.
Also I did some cleanup in GetInstance() adding some missing types, etc... could u try if it's stable with previous releases?

jonataslaw:
Great contribution, and LGTM.

lib/src/navigation/routes/bindings_interface.dart
```diff
@@ -1,7 +1,34 @@
...
 
+/// Simplifies Bindings generation from a single callback.
+/// To avoid the creation of a custom Binding instance per route.
+///
+/// Example:
+/// ```
+/// GetPage(
+///   name: '/',
+///   page: () => Home(),
+///   binding: BindingsBuilder(() => Get.put(HomeController())),
```
Nipodemos, line 17: i absolutely love that!
i was thought that having a separate file for the bindings was kinda confusing, and this will help a lot in code readability (at least for me)
thanks for that ❤️

lib/src/state_manager/rx/rx_impl.dart
```diff
@@ -15,14 +20,26 @@ class _RxImpl<T> implements RxInterface<T> {
...
+  ///
+  /// Example:
+  /// ```
+  /// var counter = 0.obs ;
+  /// counter >>= 3; // same as counter.value=3;
+  /// print(counter); // calls .toString() now
+  /// ```
+  ///
+  /// WARNING: still WIP, needs testing!
+  _RxImpl<T> operator >>(T val) {
```
Nipodemos, line 34: whaaaaaaaaaaaaaat I didn't even knew that such operator existed!

Okay i just googled and noticed that is a operator for bitwise stuff

this really needs testing and documentation, i'll make a issue about that

lib/src/state_manager/rx/rx_impl.dart
```diff
@@ -33,15 +50,24 @@ class _RxImpl<T> implements RxInterface<T> {
...
 
-  close() {
-    _subscriptions.forEach((observable, subscription) {
-      subscription.cancel();
-    });
+  @override
+  String toString() => value.toString();
+
+  /// This equality override works for _RxImpl instances and the internal values.
+  bool operator ==(o) {
```
Nipodemos, line 57: just to be sure i understood the code right: before, i could not compare Rx for equality and now i can?

so
```dart
final user = User().obs;
final user2 = User().obs;

assert(user == user2); // this would return false before and now it returns true? just to be sure

File: lib/src/state_manager/rx/rx_impl.dart [hunk 8]
```diff
@@ -286,6 +312,16 @@ class RxList<E> extends Iterable<E> implements RxInterface<List<E>> {
     subject.add(_list);
   }
 
+  /// Special override to push() element(s) in a reactive way
+  /// inside the List,
+  RxList<E> operator +(val) {
```
Nipodemos, line 317: i don't get this one, how can we use this and is it for?

lib/src/state_manager/rx/rx_impl.dart
```diff
@@ -442,24 +474,109 @@ class RxDouble extends _RxImpl<double> {
...
+    return this;
+  }
 }
 
 class RxNum extends _RxImpl<num> {
   RxNum([num initial]) {
     _value = initial;
   }
+
+  RxNum operator >>(num val) {
```
Nipodemos, line 504: i'm sorry, this question is problably very silly, but i'll ask anyway:
I case i want to actually use this operator for bitwise operations, this will interfer somehow? Sorry i am asking because I truly don't know

jonataslaw: Apparently, this operator is being used as a setter, to avoid using the .value

lib/src/navigation/routes/bindings_interface.dart
roipeker: Glad u like it, i had the same feeling about Bindings class

lib/src/state_manager/rx/rx_impl.dart
roipeker: Sure, i will add proper docs in code eventually with samples

lib/src/state_manager/rx/rx_impl.dart
roipeker: Only if u override inside User() the == and/or hashCode... but yeah it compares the inner "native" value type instead of the Rx

lib/src/state_manager/rx/rx_impl.dart
roipeker: If u wanna use that operator will be only valid for int types, mostly to shift hex values... is pretty uncommon and kinda low level. In any case if you need to do that in a RxInt just access the rx.value >> and will work on the actual int instance. For all Rx types this operator was choosen as "assignment"
So if u do
final msg="hi".obs
msg>>"goodbye"
That, even without the >>=, will assign the value and add it to the stream.

Nipodemos:
Thank you for the explantions 😄 
There is just one that you did not answered.

About overriding the plus operator in lists to assign a value
This makes possible to do like this? 
```dart
final items = RxList<String>();;
var newItem = 'foo';

items += newItem;
```

jonataslaw:
++ is a setter operator (+1)
I believe it already works in this PR as long as you don't use `final`.

items += newItem; "=" is a setter operator too

lib/src/navigation/routes/bindings_interface.dart
ghprod: Nice one @roipeker , sometimes it's a little bit over if we must create a binding file just for a controller.

Thanks

lib/src/state_manager/rx/rx_impl.dart
ghprod: So we don't need to do `user.value == user2.value` right?

roipeker: @ghprod Exactly. It compares against native type (`User()` or int, bool, etc), or the value of another Rx (`Rx<User>::value`)... so it behaves more like a "proxy" of the actual value. That's why is lacking the type in the implementation (as a **self reminder**, should be dynamic :P ).

PR #523, opened 2020-08-29
General cleanup, more docs.

roipeker:
- Rename typedefs FcBuilderFunc to InstanceBuilderCallback to be more meaningful.
- cleanup more comments.
- promote most classes in parse_route.dart (except ParseRouteTree) to be private. I still believe we should use Uri to parse the "url".
- Added _RxImpl::refresh() in rx_impl.dart and added some comments/samples.
- Some cleanup in rx_interface.dart, added more comments, missing types and move vars to the top.


jonataslaw:
LGTM, thanks for this

PR #539, opened 2020-09-01
Fixes and features

roipeker:
- Change GetPageBuilder to WidgetCallback typedef, so we can use it outside Get and the name makes more sense.
- Added isDesktop/isMobile
- Added GetPlatform (web) getters to detect the OS it runs (using navigator UA)
- Added nil extension (can be useless now), to enforce the null type safety in JS for old dart2js versions (was undefined)


jonataslaw:
Thank you for that.
I had to make some changes.
1- Obx moved, and he was trying to attach a new file, so the conflict happened.
2- While there is no folder with global settings, I prefer to keep separate typedefs for routes and Obx, because otherwise one will end up importing content from the other, and I am trying to keep each package as neutral as possible.

jonataslaw:
dart:html isn't avaible to mobile/desktop. 
I will solve this on your branch before merge

jonataslaw:
I'm going to merge and fix this in the master.
Your branch is very outdated, I was unable to make the updates without breaking everything.
After accepting the PR, I suggest that you delete your repository, and then fork again.

PR #541, opened 2020-09-01
add docs to get_view.dart

roipeker:
very simple doc reference.


jonataslaw:
LGTM! Thanks

lib/src/state_manager/simple/get_view.dart
```diff
@@ -1,6 +1,8 @@
 import 'package:flutter/widgets.dart';
 import 'package:get/src/instance/get_instance.dart';
 
+/// GetView is a great way of quickly access your Controller
+/// without having call Get.find<Controller>() yourself.
```
ghprod, line 5: @roipeker maybe you can add how to use it, something like this.

```dart
class SomeScreen extends StatelessWidget {}

// change to

class SomeScreen extends GetView<SomeController> {}
```

roipeker: Sure!, it was a quick test i was doing with @Nipodemos with the PRs yesterday.

jonataslaw:
LGTM

PR #544, opened 2020-09-01
Create internationalization.md

kai-oswald:
This is a very basic documentation to getting started with internationalization with getx.


jonataslaw:
Hi, thanks for that, this is one of the items that was missing from our roadmap.
I would just like you to include this in the main readme anyway, the documentation is not extensive, so it would not be a problem to add it to readme.md. Could you do this?

kai-oswald:
Yep, sure. I also left a link to the internationalization.md file. This can then be used to go more in-depth.

jonataslaw:
Thank's
LGTM

PR #547, opened 2020-09-01
[docs] more refactor to dependencies docs

Nipodemos:
Hello!
I've noticed that the dependency docs were kinda with duplicated docs, one for introduce the user to it, and other with the details.
I realized that it is much better to merge them together so is feels like is much more well explained then before, although i added almost nothing to it


jonataslaw:
LGTM

PR #551, opened 2020-09-02
dialog style

roipeker:
Added TextStyle to generalDialog title and message


jonataslaw:
Thank you for that.
I just made a comment about an added line that probably doesn't work, after checking this I will merge

jonataslaw:
LGTM

PR #556, opened 2020-09-03
fixes for Controller and improvements with extensions and BindingBuilder.

roipeker:
- fixes the onClose(), onInit() async override in GetxController
- added a shortcut `BindingsBuilder.put()` to register a single Controller in a view.
- fixes some doc comments
- added new extensions for `num` and `duration` to apply delayed calls.

### DIFF
--- a/lib/src/state_manager/rx/rx_core/rx_interface.dart
+++ b/lib/src/state_manager/rx/rx_core/rx_interface.dart
@@ -1,5 +1,7 @@
 import 'dart:async';
+
 import 'package:flutter/scheduler.dart';
+
 import '../rx_typedefs/rx_typedefs.dart';
 
 /// This class is the foundation for all reactive (Rx) classes that makes Get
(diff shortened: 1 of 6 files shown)


lib/src/state_manager/rx/rx_core/rx_interface.dart
```diff
@@ -78,7 +81,7 @@ abstract class DisposableInterface {
   /// Or dispose objects that can potentially create some memory leaks,
   /// like TextEditingControllers, AnimationControllers.
   /// Might be useful as well to persist some data on disk.
-  Future<void> onClose() async {}
+  void onClose() async {}
```
jonataslaw, line 84: Won't that produce an error in Get.delete?

jonataslaw: I'm going to merge it, for backwards compatibility, but I think this will generate a warning in Get.delete

Nipodemos:
What if instead of `BindingBuilder.put()` we added a simple `controller:` parameter on the GetPage? I think it would be easier to understand the purpose

jonataslaw:
Or controllers?

Nipodemos:
> Or controllers?

Well, it would be a way of split things
A place for the controllers and the bindings for other dependencies

roipeker:
Shortens some typing, is more accurate, but might create some confusions of "why both". Plus we might have to update the API in every get navigation method. And the initialization has to be a callback, otherwise you construct all your controllers in GetPages, and u will not be able to recreate the instance when the page opens again.

lib/src/state_manager/rx/rx_core/rx_interface.dart
roipeker: I think it fixes the error that we had when generating the override in the IDEs (Future<Function>). Anyway, I tested an async operation on onClose, and code execution was fine

Nipodemos:
that's true... guess people often want one good way of doing things instead of three different ways that they don't even know the difference or which one is better

Issue #558, opened 2020-09-03
Rx >>= operator should be changed to <<=


kranfix:
## Problem

The current api por a Rx object could be confused for emit a new value in a Rx implementation  
https://github.com/jonataslaw/getx/blob/d942581e47f797647ee7e95a7554a540d0ca0117/lib/src/state_manager/rx/rx_core/rx_impl.dart#L25  
 

```dart
// Currently
counter >>= 3; // Is counter injecting data to a const value?

// Should be
counter <<= 3; // A const value is injected to counter
```

Regardless of whether the feature is completed or tested, the operator should be according to its intention.

jonataslaw:
@roipeker What do you think?
How did you implement this feature, I'll let you answer that

roipeker:
@jonataslaw, @kranfix I implemented the << feature based on "votes" :D , it was marked as experimental, and since most of devs are already using `rx.value` or `rx()` ... I didn't touch it again.
I definitly agree with the `<<` operator, makes more "sense" (although still looks very strange).

kranfix:
I know that it's strange in Flutter, but as an old C++ dev, the << operator is very clear.

```C++
std::cout << "some value" << endl;
```

Where `cout` is a `stream` object.

kranfix:
On the other hand, this discussion it not about whether that feature should exist (I would like to), but also whether the operator is best meanwhile or later.

kranfix:
@roipeker If you are agree, will you update it or do you prefer me to send a PR?

PR #560, opened 2020-09-04
added json support to Rx

roipeker:
resolves the `value` (for native types) or `value.toJson()` in Rx.


Nipodemos:
closes #558

Nipodemos:
i don't get it, how should i write code that uses this?
usually i use https://app.quicktype.io/ to make a model class that converts to and from json, and Rx usually is for only one variable.
So this is useful when you make the entire class reactive?

jonataslaw:
I'm merge it up with caveats, because I'm still not sure if the "toJson" method can't collide with model classes that can use this function internally.

PR #534, opened 2020-08-31
[docs] Change "namedRoutes" for "GetPages" in docs

Nipodemos:
Oh gosh, i can't believe i have left this for so long

Until this day, the docs were still mentioning `namedRoutes` instead of `GetPages`, which is a change made months ago...

PR #561, opened 2020-09-04
fix(docs): fix page argument in GetPageRoute that needs to be a function

Nipodemos:
Hello! as @wbemanuel mentioned in [this issue comment](https://github.com/jonataslaw/getx/pull/534#issuecomment-686629270), the page argument needs to be a function that returns a widget. This PR fix it


jonataslaw:
Thank's 
LGTM

PR #564, opened 2020-09-05
fix(lint/analyser): Added analyser + effective dart

Grohden:
I discussed this briefly with @jonataslaw and he accepted the idea to put and apply `effective_dart` in this package.

I had to fix a lot of things, and tried to not break the existing api in a hard manner, but unfortunately strong mode reported a lot of type errors on RxSet, so I removed some methods from it. 

Here's a simple list of what I did:
* Reformated all sources under lib
* Fixed almost all (now) detected lint and type issues
* Added public api type annotations (fixes issues for people using strong mode)
* Broke RxSet api (appeared to be a copy from RxList with wrong methods, this needs to be rewrited)
* Broke anyone that relied on **wrong** type annotations that were omitted from this package
* Maybe broke controller close overrides (`DisposableInterface` from rx core seems to have been always returned a future close method)

This is of course open for discussion about which rules we should be using.

Note 1: I disabled some rules while I was trying to not break the existing api, but I had to break it anyway.. so I could also break it more enabling those rules (I'm talking about the `avoid_positional_boolean_parameters` and `constant_identifier_names` rules)

Note 2: 
  * Ran all tests, all passing.
  * Ran the example app, it's working
  * Applied the package on a personal app, only had to fix the `DisposableInterface` signature issue (`onClose` override should be `Future onClose`, which I had before this refactor tbh)


Grohden:
Oh, I also added some fixmes, I could fix them but I request some comments on them about how should I fix them

jonataslaw:
Thank you, this is a positive change (and a big step).
I will comment on the code as I review it, because it really was a great PR.

lib/src/instance/get_instance.dart
```diff
@@ -131,12 +144,15 @@ class GetInstance {
       }
     });
 
+    // FIXME: this will not await sequentially as probably is expected
```
Nipodemos, line 147: hello! sorry i don't understand what you meant by this, could you give a little explanation?

Nipodemos: oh i got it now, the recommended is `for in` in these cases. But why have you chose not to fix this? it would be one more breaking change in the code?

Grohden: @Nipodemos 
Here's a more visible example of the forEach vs for with await:
https://dartpad.dev/8430ebb2f2cc046a9cf2f274aea06ae4

I choose to not fix it cuz I don't know if this is the expected behavior or not (the sequential stuff)

Nipodemos: i am honestly confused for two reasons:
1 - in your dartpad, the `for in` was not sequential, but it was waiting for the time
2 - `forEach` was sequential, but didn't wait the time asked
image

i'm lost
was this expected output??

Nipodemos: this is crazy man, forEach have such an unpredicted behavior

Grohden: Oh, sorry, the sequential part refers to the array sequence, not the numerical one

forEach will take 6 seconds to finish all outputs

for in await will take the sum of all seconds to output

If I change this part of the code to be for in await it will start to take theoretically more time to execute.

Nipodemos:
what a PR
Could you list what exact methods/classes/variables that have a breaking change?

also @jonataslaw since there is so many breaking changes, according to semver now you need  to make it GetX 4.0.0 right?
It would help if you do that together with a small guide of how to adapt the changes, if this is done correctly i think that even the changes that @Grohden supressed can be added too

jonataslaw:
@Nipodemos Despite being a break update, it doesn't matter about breaking changes for the user, so it wouldn't have the version upgrade, unless any of the settings matter in exchange for code.

I reviewed it to a certain extent, and left two comments, as it is an extensive PR I will need some more time to make sure that nothing done on it can eventually break some current resource, however, as we are receiving several PRs every day, we have to have a certain urgency to merge this change, to prevent PRs from coming into conflict.

The most critical part is RxSet, which in theory should not have broken, and some things that involve futures, such as Get.delete, which has onClose which, if asynchronous, should be waited on, to avoid memory leak.

When going to onClose, he must wait, or else we can use Queue to insert exclusions in a queue (I think I even did this in a recent update, because Flutter changed the order of the disposition and forced me to have this control).
When an instance is no longer used, it receives callbacks from all sides to be excluded, both from the GetX device, from the routes, and from the route widget, and in the past even the observer used to do this. It is precisely because of this "security" that totally prevents any case of memory leak that makes us have to wait for a close-up. The closing of a Stream is asynchronous, for example, if we do not expect this, the instance will be deleted, and the method will lose the reference to the controller, and will not be able to close the stream (memory leak), so we need to wait for the close of a controller always.

Nipodemos:
well if the problem is just to add a few `Future<void>` in some methods then it won't be such a bad thing, so i agree with you.
We just need to focus in making a small guide of what change and how to adapt.

I had this error in a onClose method that I was not messing in ages, and had no clue about how to solve it. Now i know that is just add `Future<void>`, but to the rest of userts that does not follow the issues and PR this can become a nightmare

if you want i can certainly do this piece of doc, but i don't know how to find what's change in the code 😅  so if you tell me i can write it

Grohden:
Question, was RxSet working in previous versions?
RxSet was relying recursively on itself (RxList for example was using a internal List, but RxSet was "using" a internal RXSet)
Also, there's some operations that didn't make sense on a Set, like the sort one (which I couldn't find in dart's set) or the index access operator

I'm gonna review and restore/add the needed ops tomorrow/today on RxSet

Sorry for the big PR, I was almost done when I thought about disabling the 80 column limit rule.. I still can revert those if you wish, so this PR will have less diffs

Grohden:
@jonataslaw @Nipodemos 

What is probably broken:

* `DisposableInterface` `onClose async {}` is declared as `Future onClose() async {}`:

Why:
There's a piece of code at `get_instance.dart` (329):
```dart
     if (i is DisposableInterface) {
        await i.onClose();
        GetConfig.log('"$newKey" onClose() called');
      }
```
`await` can work with non futures, but I assumed that the onClose was intended to be a future, so I used `Future` instead of `void`, we can change to `void` and remove the await at that line if that's what should be done

* `GetUtils` `hasMatch` changed it's signature:

Before: `static bool hasMatch(String value, Pattern pattern)`
Now:  `static bool hasMatch(String value, String pattern)`

Why: `RegExp(pattern)` signature matches only `String` and not `Pattern`, I'm not sure if I a call .toString on the `pattern` would work. I found easier to just follow the `RegExp` type signature


I had to fix some type signatures that were wrong and/or accepted dynamic generics on:

* `RxMap`

changed `void addListener(Stream rxGetx)` to `void addListener(Stream<Map<K, V>> rxGetX)`


* `RxList`

Some signatures accepted a `Iterable<E>` and assigned it to a `List<E>` without calling `.toList`, I've opted to change 
those signatures to `List<E>` instead of using a .toList call because implicit conversions are not a good idea.

What changed:
* `void bindStream(Stream<Iterable<E>> stream)` fixed to `void bindStream(Stream<List<E>> stream)`
* `set value(Iterable<E> val)` fixed to `set value(List<E> val)`
* `close` added the `void` return signature
* `operator []=(int index, E val)` added the `void` return signature

Other type signatures that were added to probably public apis:
* `TaskManager` `notify` added `void` as return, previously omitted.
* `ValueBuilder` was omitting the `T` requirement from `ValueBuilderBuilder`, now it only accepts `ValueBuilderBuilder<T>`, previously it accepted  `ValueBuilderBuilder<dynamic>`


This comment doesn't cover RxSet yet

Grohden:
`RxSet`

I've removed these:

```
  void insert(int index, E item) {
    _set.insert(index, item); // doesn't exist on set.
    subject.add(_set);
  }

  void insertAll(int index, Iterable<E> iterable) {
    _set.insertAll(index, iterable);  // doesn't exist on set.
    subject.add(_set);
  }

   E removeAt(int index) {
    E item = _set.removeAt(index);
    subject.add(_set);
    return item;
  }

  E removeLast() {
    E item = _set.removeLast();
    subject.add(_set);
    return item;
  }

  void removeRange(int start, int end) {
    _set.removeRange(start, end);
    subject.add(_set);
  }

  void sort([int compare(E a, E b)]) {
    _set.sort();
    subject.add(_set);
  }

```

Dart type checker probably accepted these as correct because  `_set` was declared as `RxSet<T> _set = Set<T>()` (which was wrong)

Their signatures where basically the same of the RxList and didn't match type checking nor would be expected to work because set doesn't have `insert` and the remove ones.

I see that basically all the correct set methods are already overridden and reactive, so basically removing the wrong methods was the only action that needed to be taken 

Also, there's a lot of refs in doc comments for 'lists' instead of set

Grohden:
Just thinking now, about the Future changes, I could easily solve that with a FutureOr 🤔 it probably wouldn't require people to mess with old onClose signatures...
Gonna make a test later.

jonataslaw:
Sorry for the delay, I was only able to review it now.
I think you are right about RxSet.
The nonexistent methods need to be removed.

Nipodemos:
> Question, was RxSet working in previous versions?
> RxSet was relying recursively on itself (RxList for example is using a internal List, but RxSet was "using" a internal RXSet)
> Also, there's some operations that didn't make sense on a Set, like the sort one (which I couldn't find in dart's set) or the index access/assignment operator

RxSet was implemented just a few days ago, so it has the chance that nobody used it (at least no one made a issue about that)
so that's why

jonataslaw:
> @jonataslaw @Nipodemos
> 
> What is probably broken:
> 
> * `DisposableInterface` `onClose async {}` is declared as `Future onClose() async {}`:

I'm still not sure what to do about it, but this API has been breaking, because many users have used "void" as a type for a long time. I don't know if FutureOr would cause another API change, that we will have to investigate.

> Why:
> There's a piece of code at `get_instance.dart` (329):
> 
> ```dart
>      if (i is DisposableInterface) {
>         await i.onClose();
>         GetConfig.log('"$newKey" onClose() called');
>       }
> ```
> 
> `await` can work with non futures, but I assumed that the onClose was intended to be a future, so I used `Future` instead of `void`, we can change to `void` and remove the await at that line if that's what should be done
>
> * `GetUtils` `hasMatch` changed it's signature:
> 
> Before: `static bool hasMatch(String value, Pattern pattern)`
> Now: `static bool hasMatch(String value, String pattern)`
> 
> Why: `RegExp(pattern)` signature matches only `String` and not `Pattern`, I'm not sure if I a call .toString on the `pattern` would work. I found easier to just follow the `RegExp` type signature
> 
> I had to fix some type signatures that were wrong and/or accepted dynamic generics on:
> 
> * `RxMap`
> 
> changed `void addListener(Stream rxGetx)` to `void addListener(Stream<Map<K, V>> rxGetX)`


It's ok

> * `RxList`
> 
> Some signatures accepted a `Iterable<E>` and assigned it to a `List<E>` without calling `.toList`, I've opted to change
> those signatures to `List<E>` instead of using a .toList call because implicit conversions are not a good idea.

The Dart List Api uses Iterable, because there are other Lists that are of different types, such as:
List <String> list = <String>['Jonny', 'Gabriel'];
List <String> list2 = = <String>[];

list2.addAll(list.reversed);

Thus, any addition to a List must be an Iterable, not a List.
I think this part is going to need a regression.

> What changed:
> 
> * `void bindStream(Stream<Iterable<E>> stream)` fixed to `void bindStream(Stream<List<E>> stream)`

This will remove BindStream compatibility for other types of Iterables.

> * `set value(Iterable<E> val)` fixed to `set value(List<E> val)`

This will probably prevent someone from making a setter from a Reverse List to a normal list

> * `close` added the `void` return signature
> * `operator []=(int index, E val)` added the `void` return signature

It's ok

> Other type signatures that were added to probably public apis:
> 
> * `TaskManager` `notify` added `void` as return, previously omitted.

It's ok

> * `ValueBuilder` was omitting the `T` requirement from `ValueBuilderBuilder`, now it only accepts `ValueBuilderBuilder<T>`, previously it accepted  `ValueBuilderBuilder<dynamic>`

It's ok

> This comment doesn't cover RxSet yet

Grohden:
OK, can confirm that using `FutureOr` make things simple to deal with from a migration point of view

Using `FutureOr onClose async {}` on `DisposableInterface` makes these override signatures acceptable:
```dart
  @override
  Future onClose()

  @override
  FutureOr onClose()

  @override
  void onClose()
```

> The Dart List Api uses Iterable, because there are other Lists that are of different types, such as:
> List list = ['Jonny', 'Gabriel'];
> List list2 = = [];
>
> list2.addAll(list.reversed);

That is still working, list signatures will be the same as the overridden ones from dart std.

Iterable vs List changes are on overrides that come from `RxInterface<List<E>>`, since the subject is basically `List<E>`, sets and listeners will be expected to be dealing with a List instead of a iterable

> This will remove BindStream compatibility for other types of Iterables.

For example, using the iterable signature on this method:
```dart
 void bindStream(Stream<Iterable<E>> stream) =>
      stream.listen((va) => value = va);
```
Causes a type check error on the assignment `value = va` (`A value of type 'Iterable<E>' can't be assigned to a variable of type 'List<E>'.`)

That is, `Iterable` is not a `List`, but a `List` is a `Iterable`, someone could make a `class Foo<E> extends Iterable<E>` and with that signature the compiler would accept a bind stream with `Stream<Foo<E>>`, but we would get a runtime error once the assignment of a `Foo<E>` to a `List<E>` is made.

> This will probably prevent someone from making a setter from a Reverse List to a normal list

Same problem

This can be solved in some other ways:
* Implicit `.toList` calls
* Change `RxInterface<List<E>>` to `RxInterface<Iterable<E>>`, which would cause:
  * My signatures changes to be reverted (`bindStream`, `set value`, `addListener`)
  * `StreamController<List<E>> subject` would change to `StreamController<Iterable<E>> subject`
  * `Map<Stream<List<E>>, StreamSubscription>` would change to `Map<Stream<Iterable<E>>, StreamSubscription>`
  *  `Stream<List<E>> get stream` would change to `Stream<Iterable<E>> get stream`
      * Original signature could be keept with a `subject.stream.map((event) => event.toList());`
  * `set value` would need to call `.toList` at the `_list` assignment

lib/src/utils/regex/get_utils.dart
```diff
@@ -66,7 +71,9 @@ class GetUtils {
 
   /// Checks if string is an audio file.
   static bool isAudio(String filePath) {
-    String ext = filePath.toLowerCase();
+    final ext = filePath.toLowerCase();
+
+    // FIXME: why all those ()?
```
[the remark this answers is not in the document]
Grohden: ok, gonna remove them and apply dart formatter

lib/src/instance/get_instance.dart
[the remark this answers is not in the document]
Grohden: ok, gonna replace with a for loop with await inside

example/test/test.dart
```diff
@@ -1,26 +1,30 @@
...
   Future<CasesModel> getCases() async {
     await Future.delayed(Duration(milliseconds: 100));
-    return Random().nextBool()
-        ? CasesModel(
-            global: Global(totalDeaths: 100, totalConfirmed: 200),
-          )
-        : Future.error('error');
+
+    if (Random().nextBool()) {
+      CasesModel(
```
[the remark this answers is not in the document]
Grohden: oh boy, thanks for noticing that!

jonataslaw:
Okay, you're right.
I was just thinking of creating a mixin, or an abstract class to take advantage of the mandatory methods of an Iterable, and use List, and Set in a more granular way, but that is not so necessary at the moment.
Do you have any FixMe pending yet? If not, I'll merge.
Thank you for this contribution

Grohden:
@jonataslaw, in the future, creating a RxIterable and reuse it at set and list is a great idea!
I'm gonna push those required fixes in the next hour

Grohden:
@jonataslaw added 5 atomic fix/refactor commits with fixes

I also see that get will need to add `flutter analyze` into it's PR pipeline in the near future, if you wish I could open a PR for that

jonataslaw:
> @jonataslaw added 5 atomic fix/refactor commits with fixes
> 
> I also see that get will need to add `flutter analyze` into it's PR pipeline in the near future, if you wish I could open a PR for that

That would be really good, 
and thank you very much, this was a great contribution, and I think it can already be merged.

Issue #507, opened 2020-08-26
`GetxController`의 해당 기능이 3.4.1 이후로 사라졌습니다.


idompolo:
In GetxController
   Disposer addListenerId (String key, StateSetter listener) {
     _updatersIds [key] = listener;
     return () => _updatersIds.remove (key);
   }

Until version 3.4.1, all duplicate GetBuilders were updated even though the key values were duplicated.
Later versions only update the last created GetBuilder.

I am very in trouble now
I did a lot of coding with version 3.4.1.
It is a situation that needs change.
Please restore.

Nipodemos:
but the Ids should be unique
But to be honest that's a clever idea to manage what rebuilds without giving an id for each and every GetBuilder
@jonataslaw have you changed something that made this issue happen?

jonataslaw:
This would be possible if instead of Map, we had a Map<String, List <StateSetter>>.
This would bring a greater overhead, as the Ids would go through 2 loops.
This change was made in order to improve performance and build time, although I think it is a valid suggestion, I am not sure that this could not bring performance problems.
I'm going to do tests to make sure there is a less expensive way to do this.

roipeker:
I understand the use case and _id_ maybe should be _uniqueId_. Otherwise is like a _groupId_. I agree with @jonataslaw and the overhead, it will apply to every GetxController.update()

I will try a workaround, although will increase the logic complexity of the updates, but avoid the forced double loop.

roipeker:
Just pushed a PR for this issue (+ other improvements). Although is a big change, and i would like to have some feedback from @jonataslaw.
I ran my tests, and seems to be working fine, and is the most "performant" way i came up with (following @jonataslaw  reference that `HashMap` is faster to loop than `LinkedList`).
The update should be transparent to developers, no external API changed.

Just in case, [here is a quick code](https://gist.github.com/roipeker/f34654ca35b80f5bf34736bd865f6358.) to test the PR.

Nipodemos:
This bug should also be fixed with this merged PR
@idompolo can you test it for us? but you need to wait for @jonataslaw to release the new version with this fix

If the problem persists, feel free to reopen this issue or create another one

idompolo:
> 이 버그는 [# 565](https://github.com/jonataslaw/getx/pull/565) 로 수정되어야합니다.
[...]

Issue #465, opened 2020-08-12
Transition animation curve behaviour


gary6lin:
Hi,

I have set the 'curve' property in GetPageRoute and GetPageRoute and GetPage but it seems not working as expected.

So I dig into the source code and found that the curve is not used for most of the transitions. Is this the intended behaviour or something that has been missed out?

In addition to this, it would be great if we can set the transition curve globally.

Thanks :)

jonataslaw:
Yes, most transitions do not use this property.
However, you can create CustomTransitions with it.

gary6lin:
> Yes, most transitions do not use this property.
> However, you can create CustomTransitions with it.

But if I set the CustomTransitions globally on the GetMaterialApp, I won't be able to use the preset Transitions because the buildTransition() checks for the customTransition property first. I will have to set the customTransition for each route if I don't want to use the global default.

Are we able to improve this?

Thanks

roipeker:
Fixed in this [PR](https://github.com/jonataslaw/getx/pull/565/commits/a7d193b04431c82e5bafb9331275a6880229e794)

Issue #549, opened 2020-09-01
Debounce not working properly


JCKodel:
**Describe the bug**
Debouncer is not working properly.
I have a `TextEditingController` with an listener to write changes to disk:

```dart
@override
void initState() {
  super.initState();
  _displayNameTextEditingController = TextEditingController(text: FirebaseAuth.instance.currentUser.displayName);
  _displayNameTextEditingController.addListener(_onDisplayNameChanged);
}

void _onDisplayNameChanged() {
  final userController = Get.find<UserController>();

  print("Text has changed");

  if (userController.currentUser.value.displayName == _displayNameTextEditingController.value.text) {
    print("But it is equal to the model value, so ignoring it...");
    return;
  }

  userController.updateUser((u) => u.displayName = _displayNameTextEditingController.value.text);
}
```

On `updateUser`, I save it to the disk and update the firebase user (a remote call):

```dart
Future<void> updateUser(void Function(User) updater) async {
  currentUser.update(updater);

  debounce(currentUser, (value) async {
    final fbUser = fba.FirebaseAuth.instance.currentUser;

    await userStore.save(currentUser.value, currentUser.value.id);
    await fbUser.updateProfile(displayName: currentUser.value.displayName, photoURL: currentUser.value.photoURL);
  });
}
```

The `userStore.save` has a log starting with 🟧 (meaning: the data was actually written to disk).

This is my log after 1 keystroke:
```
I/flutter (13017): Text has changed
I/flutter (13017): But it is equal to the model value, so ignoring it...
I/flutter (13017): Text has changed
```

Focus triggers the TextEditingController (but the first one is ignored because the text itself didn't change).
Then, a save is triggered (but the bounce just freezes, no data is actually written to the disk).

Now, clearing the log and pressing another key:
```
I/flutter (13017): Text has changed
I/flutter (13017): Text has changed
I/flutter (13017): But it is equal to the model value, so ignoring it...
I/flutter (13017): 🟧 user::4FuJiTZREfYhNHylSCX8rShpIei1
I/flutter (13017): 🟧 user::4FuJiTZREfYhNHylSCX8rShpIei1
I/BiChannelGoogleApi(13017): [FirebaseAuth: ] getGoogleApiForMethod() returned Gms: com.google.firebase.auth.api.internal.zzaq@7167582
[...]

PR #565, opened 2020-09-05
improve docs, cleanup workers, added samples, fixed rx operators

roipeker:
- Commented code for #549 
- Improved docs
- cleanup workers (and added docs and  added samples)
- fixed Rx operators
- simplified RxInt / RxDouble operators with base abstract class.


lib/src/state_manager/rx/rx_workers/rx_workers.dart
```diff
@@ -1,115 +1,195 @@
...
+///  class _CountController extends GetxController {
+///   final count = 0.obs;
+///   Worker worker;
+///
+///   @override
+///   Future<void> onInit() async {
+///     worker = once(count, (value) {
+///       print("counter reached $value before 3 seconds.");
+///     }, condition: () => count() > 2);
+///     3.delay(worker.dispose);
```
Nipodemos, line 88: What is this syntax that i don't know?
dart has that by default?

jonataslaw: He added the duration and delay extension in the last update, it is not from dart, but it is included in Getx

roipeker: Is just a utility for `num`. I've got tired of `await Future.delayed(Duration( seconds: 1 ));`, so the extension go for `num` and `Duration` (although, obviously you can't make it `const`) :

Duration:
`await 2.minutes.delay()` = `await 0.5.seconds.delay()`.
`DateTime.now().add(4.days);`

num: (the `delay()` always resolves as "seconds").
`await 0.3.delay();`

Nipodemos:
i guess you wanted to make that last commit to be a separated pull request, but it is together with this one
anyway, nice things to add!

lib/src/navigation/routes/bindings_interface.dart
```diff
@@ -16,14 +18,20 @@ abstract class Bindings {
 /// GetPage(
 ///   name: '/',
 ///   page: () => Home(),
-///   binding: BindingsBuilder(() => Get.put(HomeController())),
+///   // This might cause you an error.
+///   // binding: BindingsBuilder(() => Get.put(HomeController())),
+///   binding: BindingsBuilder(() { Get.put(HomeController(); })),
```
Nipodemos, line 23: why this may cause errors?

jonataslaw: Dart has been having some problems with arrow functions recently.
I also experienced errors using arrow functions in the last few days, which makes no sense, but in the dev version of Flutter (which points to the dev version of dart) this is happening frequently.
I don't know if this needs to be documented, as this problem will certainly be fixed in the dart

Nipodemos: this is indeed weird, and I agree with you, although is a nice thing, this does not need to be documented here

roipeker: I can remove the comment, but we should wait til this gets solved. As long as u use lazyPut() or the new factory constructor `BindingsBuilder.put` for 1 reference, you should be fine.

lib/src/state_manager/simple/get_state.dart
```diff
@@ -68,15 +139,20 @@ class GetBuilder<T extends GetxController> extends StatefulWidget {
...
 }
 
-class _GetBuilderState<T extends GetxController> extends State<GetBuilder<T>> {
+class _GetBuilderState<T extends GetxController> extends State<GetBuilder<T>>
+    with GetStateUpdaterMixin {
   GetxController controller;
   bool isCreator = false;
-  final HashSet<Disposer> disposers = HashSet<Disposer>();
-  Disposer remove;
+
```
jonataslaw, line 151: This was a remnant of the "SimpleBuilder" test, it can be removed, thanks for taking this.

lib/src/state_manager/simple/get_state.dart
```diff
@@ -8,39 +8,109 @@ import 'package:get/state_manager.dart';
...
-  // void removeListener(StateSetter listener) {
-  //   _updaters.remove(listener);
-  // }
+//  VoidCallback addListenerId(String key, StateSetter listener) {//<old>
+  VoidCallback addListenerId(String key, GetStateUpdate listener) {
+//    _printCurrentIds();
+    if (_updatersIds.containsKey(key)) {
+      final _innerKey = _groupIdCount++;
+//      final _ref = _updatersGroupIds[key] ??= HashMap<int, StateSetter>();//<old>
+      final _ref = _updatersGroupIds[key] ??= HashMap<int, GetStateUpdate>();
```
jonataslaw, line 77: In my tests the update with ID is slower than before.
Just about the Iterables speed issue:
HashSet is the fastest of all in any situation.
LinkedList is fast for a few items, and HashMap starts to take advantage from the tenth item.
I believe that a HashMap<String, HashSet> could reduce that difference dramatically, at least in theory.

jonataslaw (commented):
I made some considerations, and resolving the HashMap issue, this can be merged

lib/src/state_manager/simple/get_state.dart
roipeker: @jonataslaw . Ok, if that's the only change, i will push it soon.

Nipodemos:
What changed with these commits:
#### Fixes transitions and curves arguments in navigation methods
- renamed and added defaults transition duration and types in "GetInterfase"
- added missing parameters in `Get.to` / `Get.offAll`,  fixes #570
- added optional `transitionDuration` and `transitionCurve` to `Get.dialog()`

#### optimize GetBuilder "group id" using `HashSet`

- `Changed HashMap<int,GetStateUpdate>` to `HashSet<GetStateUpdate>`.
 Reference #565 (comment)
-  Added a internal VoidCallback in `GetStateUpdaterMixin::getUpdate`, to see if it reduces overhead of anonym function in `setState()` (considering GC).

#### fixed GetPages.curve

- fixes #465
- iOS has an exception to the rule, when no Curve is defined, it doesn't use linearTransition in `CupertinoPages` (`Transition.cupertino/native/default`) and fullscreen dialogs so it applies the Curve, or use the default one.
It might be smarter, maybe, to also enforce the `animation` duration to stay in platforms defaults.
- had to change `CupertinoTransitions.buildTransitions` params in order to make the check and keep the code change small.

jonataslaw:
> What changed with these commits:
> 
> #### Fixes transitions and curves arguments in navigation methods
> * renamed and added defaults transition duration and types in "GetInterfase"
> * added missing parameters in `Get.to` / `Get.offAll`,  fixes #570
> * added optional `transitionDuration` and `transitionCurve` to `Get.dialog()`
> 
> #### optimize GetBuilder "group id" using `HashSet`
> * `Changed HashMap<int,GetStateUpdate>` to `HashSet<GetStateUpdate>`.
>   Reference #565 (comment)
> * Added a internal VoidCallback in `GetStateUpdaterMixin::getUpdate`, to see if it reduces overhead of anonym function in `setState()` (considering GC).
> 
> #### fixed GetPages.curve
> * fixes #465
> * iOS has an exception to the rule, when no Curve is defined, it doesn't use linearTransition in `CupertinoPages` (`Transition.cupertino/native/default`) and fullscreen dialogs so it applies the Curve, or use the default one.
>   It might be smarter, maybe, to also enforce the `animation` duration to stay in platforms defaults.
> * had to change `CupertinoTransitions.buildTransitions` params in order to make the check and keep the code change small.

thanks after that i would upload a version, and this description was very useful for changeLog

jonataslaw:
Thank you for this great contribution!
I just resolved her conflicts with the other effective dart PR.
LGTM.

Nipodemos:
just to make it more complete @roipeker also mentioned this in his commit descriptions:

- improved code docs for `BindingsBuilder` clarifying the issue with fat arrows and Get.put()
- major change in `GetxController`, `GetBuilder`, `SimpleBuilder` to improve docs, code cleanup (left a "clean" mess with the previous code to revert if needed)
- new `GetStateUpdaterMixin` and `GetStateUpdate` in favour of `StateSetter` (big change).
- Fixes issue #507, quick code for testing multiple GetBuilder ids can be found here: https://gist.github.com/roipeker/f34654ca35b80f5bf34736bd865f6358.
Warning: is a big change, but I did my tests and seems to be working fine. Please, test performance and potential bugs.

Grohden:
@Nipodemos @jonataslaw @roipeker 
> simplified RxInt / RxDouble operators with base abstract class.

Is this part working as expected?

I ask this because this abstraction broke my old code:

`abstract class _BaseRxNum<T> extends _RxImpl<num> {`

This makes a `int` become a `num`, so using something simple like:

```dart
   // ctrl
   final currentRoute = 0.obs;

   // view
   final index = ctrl.currentRoute.value;
   final route = _homeItems[index];
```

Reports me this:
`error: The argument type 'num' can't be assigned to the parameter type 'int'. (argument_type_not_assignable at [app] ...`

roipeker:
> @Nipodemos @jonataslaw @roipeker
> 
> > simplified RxInt / RxDouble operators with base abstract class.
> 
> Is this part working as expected?
> 
> I ask this because this abstraction broke my old code:
> 
> `abstract class _BaseRxNum<T> extends _RxImpl<num> {`
> 
> This makes a `int` become a `num`, so using something simple like:
> 
> ```dart
>    // ctrl
>    final currentRoute = 0.obs;
> 
>    // view
>    final index = ctrl.currentRoute.value;
>    final route = _homeItems[index];
> ```
> 
> Reports me this:
> `error: The argument type 'num' can't be assigned to the parameter type 'int'. (argument_type_not_assignable at [app] ...`

That's very weird. `currentRoute` should be detected as `RxInt`, please try:
```dart
print( "rx: ${currentRoute.runtimeType}, value: ${currentRoute.runtimeType}");
```
Please @Grohden, check if this snippet gives you an error as well (I guess is similar to what you are doing):

```dart
import 'package:flutter/material.dart';

void main() {
  Get.lazyPut(() => _HomeController());
  runApp(GetMaterialApp(home: _Home()));
}

class _HomeController extends GetxController {
  final currentRoute = 0.obs;
  void nextRoute() => currentRoute + 1;
}

class _Home extends GetView<_HomeController> {
  final _homeItems = [12, 13, 14];

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: [
            Obx(() {
              final index = controller.currentRoute.value % _homeItems.length;
              final route = _homeItems[index];
              return Text('Route is $route');
            }),
            FlatButton(
                onPressed: controller.nextRoute,
                child: Icon(Icons.navigate_next)),
          ],
        ),
      ),
    );
  }
}
```

Grohden:
@roipeker it is a `RxInt`, the problem is here: `_RxImpl<num>` your subject is a `num`, not a `int` nor a `double`

A way to solve it is to use `T extends num`, but you will need to rewrite some of your abstracted methods down in the hierarchy, because some of them will not generate the same `T` that is expected (eg: a/b -> double, but the T is expected to be a int)

And btw, yes, your snippet gives me the same error

roipeker:
I should use a more aggressive linter rules I guess. Thanks for the reporting!

jonataslaw:
I just tested this and the problem actually occurs.
Thanks for alerting this now @Grohden

Grohden:
@roipeker @jonataslaw now that we have strong mode enabled, we just need to write this cases in the example app... or add a test, analyzer will warn us in the future as long as we have the case in the code

PR #573, opened 2020-09-07
rx list sort method compare param passed

Antoraro:


Nipodemos (approved):
nice catch!

jonataslaw:
Thanks for your contributation
LGTM

PR #574, opened 2020-09-08
subtle Rx improvements and docs.

roipeker:
- small comment fix (https://github.com/jonataslaw/getx/pull/560#commitcomment-42045326)
- added `RxBool.toggle()`as an easy shortcut for switching true/false values.
- added `_RxImp.nil()` to easily set the value to `null` (`nil` comes from LISP :P) seems to be a useful sometimes, for instance if using `RxString` in `TextField.decoration.errorText` is the only way to avoid the "error state".
- adds missing docs to Rx classes.

### DIFF
--- a/lib/src/state_manager/rx/rx_core/rx_impl.dart
+++ b/lib/src/state_manager/rx/rx_core/rx_impl.dart
@@ -3,8 +3,12 @@ import 'dart:collection';
 
 import '../rx_core/rx_interface.dart';
 
+
+/// global object that registers against `GetX` and `Obx`, and allows the reactivity
+/// of those `Widgets` and Rx values.
 RxInterface getObs;
 
+/// Base Rx class that manages all the stream logic for any Type.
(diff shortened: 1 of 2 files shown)


lib/src/state_manager/rx/rx_core/rx_impl.dart
```diff
@@ -138,13 +161,17 @@ class _RxImpl<T> implements RxInterface<T> {
 
   bool firstRebuild = true;
 
+
+  /// Updates the [value] and adds it to the stream, updating the observer
+  /// Widget, only it's different from the previous value.
```
Nipodemos, line 166:
```suggestion
  /// Widget, only if it's different from the previous value.
```

Nipodemos (approved):
seems fine to me! thank you for the contirbution

jonataslaw:
LGTM, thanks for your contribution

PR #576, opened 2020-09-08
new README docs section

### LINKED ISSUE #470: Add docs: GetxService, GetView and GetWidget
This is just a temporary roadmap about the documentation, which I checked now and there are some undocumented points that I think are essential for productivity with Getx.

- GetxService (And the respective difference between Get.lazyPut's fenix mode, and Get.put's permanent mode).
- GetView (StatelessWidget with easy controller acess);
- GetWidget (StatelessWidget with easy controller acess and cache, perfect for use with Get.create());

roipeker:
- fixes #470
- added main README section docs called "Useful tips" where i mentioned some stuffs..

### DIFF
--- a/README.md
+++ b/README.md
@@ -786,7 +786,7 @@ void initServices() async {
   /// Here is where you put get_storage, hive, shared_pref initialization.
   /// or moor connection, or whatever that's async.
   await Get.putAsync(() => DbService().init());
-  await Get.put(SettingsService()).init();
+  await Get.putAsync(SettingsService()).init();
   print('All services started...');
 }
 
--- a/README.md
+++ b/README.md
@@ -355,23 +355,25 @@ return GetMaterialApp(
 
 ## Change Theme
 
-Please do not use any higher level widget than GetMaterialApp in order to update it. This can trigger duplicate keys. A lot of people are used to the prehistoric approach of creating a "ThemeProvider" widget just to change the theme of your app, and this is definitely NOT necessary with Get.
+Please do not use any higher level widget than `GetMaterialApp` in order to update it. This can trigger duplicate keys. A lot of people are used to the prehistoric approach of creating a "ThemeProvider" widget just to change the theme of your app, and this is definitely NOT necessary with **GetX™**.
 
-You can create your custom theme and simply add it within Get.changeTheme without any boilerplate for that:
+You can create your custom theme and simply add it within `Get.changeTheme` without any boilerplate for that:
 
 ```dart
 Get.changeTheme(ThemeData.light());
 ```
 
-If you want to create something like a button that changes the theme with onTap, you can combine two Get APIs for that, the api that checks if the dark theme is being used, and the theme change API, you can just put this within an onPressed:
+If you want to create something like a button that changes the Theme in `onTap`, you can combine two **GetX™** APIs for that:
+- The api that checks if the dark `Theme` is being used. 
+- And the `Theme` Change API, you can just put this within an `onPressed`:
 
 ```dart
 Get.changeTheme(Get.isDarkMode? ThemeData.light(): ThemeData.dark());
 ```
 
-When darkmode is activated, it will switch to the light theme, and when the light theme is activated, it will change to dark.
+When `.darkmode` is activated, it will switch to the _light theme_, and when the _light theme_ becomes active, it will change to _dark theme_.
 
-If you want to know in depth how to change the theme, you can follow this tutorial on Medium that even teaches the persistence of the theme using Get:
+If you want to know in depth how to change the Theme, you can follow this tutorial on Medium which even teaches the persistence of the theme using **GetX™**:
 
 - [Dynamic Themes in 3 lines using Get](https://medium.com/swlh/flutter-dynamic-themes-in-3-lines-c3b375f292e3) - Tutorial by [Rod Brown](https://github.com/RodBr).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -355,23 +355,25 @@ return GetMaterialApp(
 
 ## Change Theme
 
-Please do not use any higher level widget than GetMaterialApp in order to update it. This can trigger duplicate keys. A lot of people are used to the prehistoric approach of creating a "ThemeProvider" widget just to change the theme of your app, and this is definitely NOT necessary with Get.
+Please do not use any higher level widget than `GetMaterialApp` in order to update it. This can trigger duplicate keys. A lot of people are used to the prehistoric approach of creating a "ThemeProvider" widget just to change the theme of your app, and this is definitely NOT necessary with **GetX™**.
```
Nipodemos, line 358: prehistoric? aren't this the usual way of doing it without GetX?

README.md
```diff
@@ -578,8 +596,8 @@ ValueBuilder<bool>(
 ),
 ```
 
-#### ObxValue
-Similar to ValueBuilder, but this is the Reactive version, you pass a Rx instance (remember the magical .obs?) and 
+####ObxValue
```
Nipodemos, line 599:
```suggestion
#### ObxValue
```

README.md
```diff
@@ -578,8 +596,8 @@ ValueBuilder<bool>(
...
+Similar to [`ValueBuilder`](#valuebuilder), but this is the Reactive version, you pass a Rx instance (remember the magical .obs?) and
```
Nipodemos, line 600: Nice idea man, put links in between the docs helps a lot

README.md
```diff
@@ -591,6 +609,214 @@ ObxValue((data) => Switch(
...
+You have to access the real `value` inside the _observable_: 
+The most "used way" is `.value`, but, did you know that you can also use...
+
+```dart
+final name = 'GetX'.obs;
+// only "updates" the stream, if the value is different from the current one. 
+name.value = 'Hey';
+
+// this weird (and kinda cool) assignment, updates the stream no matter what
+// it takes nulls, or same value... but rebuilds the observers.
```
Nipodemos, line 636: oh so assigning like that will rebuild even if the value is the same? This kinda break one of the principles of reactive state manager: only rerender when necessary.
I think this operator should respect this principle too

stefandevo:
Awesome work guys!!

README.md
```diff
@@ -591,6 +609,214 @@ ObxValue((data) => Switch(
...
+final flag = false.obs;
+
+// switches the value between true/false
+flag.toggle();
+
+
+/// all types:
+
+// Sets the `value` to null.
+flag.nil();
```
Nipodemos, line 670: although i can understand that nil it null, this is a word that does not exist in dart, and people that never had contact with a language that used this may be confused.

The ideal in my opinion is `flag.setNull()`

i know this is not the PR to say this but i only thought about that now 😅

README.md
```diff
@@ -591,6 +609,214 @@ ObxValue((data) => Switch(
...
+abc[3]; // like Lists, reads the index 3.
+
+
+// equality works with the Rx and the value, but hashCode is always taken from the value
+final number = 12.obs;
+print( number == 12 ); // prints > true
+
+/// Custom Rx Models:
+
+// toJson(), toString() are deffered to the child, so you can implement override them, and print() the observable directly.
```
Nipodemos, line 692:
```suggestion
// toJson(), toString() are deferred to the child, so you can implement override on them, and print() the observable directly. 
```

README.md
```diff
@@ -591,6 +609,214 @@ ObxValue((data) => Switch(
...
+/// Is a smart move to make your Services intiialize before you run the Flutter app.
+/// as you can control the execution flow (maybe you need to load some Theme configuration, 
+/// apiKey, language defined by the User... so load SettingService before running ApiService.
+/// so GetMaterialApp() doesnt have to rebuild, and takes the values directly.    
+void initServices() async {
+  print('starting services ...');
+  /// Here is where you put get_storage, hive, shared_pref initialization.
+  /// or moor connection, or whatever that's async.
+  await Get.putAsync(() => DbService().init());
+  await Get.put(SettingsService()).init();
```
Nipodemos, line 789:
```suggestion
  await Get.putAsync(SettingsService()).init();
```

README.md
```diff
@@ -591,6 +609,214 @@ ObxValue((data) => Switch(
...
+    print('$runtimeType delays 2 sec');
+    await 2.delay();
+    print('$runtimeType ready!');
+    return this;
+  }
+}
+
+class SettingsService extends GetxService {
+  void init() async {
+    print('$runtimeType delays 2 sec');
```
Nipodemos, line 804:
```suggestion
    print('$runtimeType delays 1 sec');
```

README.md
```diff
@@ -591,6 +609,214 @@ ObxValue((data) => Switch(
...
+```
+
+The only way to actually delete a `GetxService`, is with `Get.reset()` which is like a
+"Hot Reboot" of your app. So remember, if you need absolute persistance of a class instance during the
+lifetime of your app, use `GetxService`. 
+  
+
+ 
+
+
```
Nipodemos, line 819: the markdown ignores by default more than one line, but even so, it's a good practice to not let more than 1 blank line

README.md
roipeker: I like short methods names haha. And having LISP nil had a pun intended with flutter. But anyway, open to suggestions, whatever the name is a must to have such utility to initialize with null

final count=0.obs..setNull();

// looks much cleaner no?
final bla=0.obs..nil();

README.md
Nipodemos: @jonataslaw

README.md
Nipodemos: i don't get the joke 😅 

well, i don't think it cleaner to be honest 😅 , but that is only my opinion

README.md
stefandevo: Just for my understanding, `Get` needs to be replaced with `GetX™` everywhere?
Because there are still some places in this PR where it is `Get`

Nipodemos: @stefandevo the ideal is to replace everywhere

Nipodemos (approved):
good enough for me

jonataslaw:
LGTM

PR #578, opened 2020-09-08
Get.delete update and some missing docs

roipeker:
- Added `Get.delete(force:false)` which was missing to manually remove permanent instances with Get.put().
- added missing doc comment to `Get.isPrepared()`


jonataslaw:
LGTM, thank's!

PR #579, opened 2020-09-08
chore: add analyzer to PR check pipeline

Grohden:
Now with strong mode enabled, we can properly validate PRs code styles, this PR adds the `flutter analyze` to the pr check pipeline

This was not simple as I would expected it to be because of this https://github.com/flutter/flutter/issues/62591, but I could avoid it by placing all `effective_dart.1.2.0.yaml` rules at getx `analisis_options.yaml` instead of using the include directive.

I've also, fixed new lint issues


jonataslaw:
Thank you for that!
LGTM

PR #589, opened 2020-09-10
Fix for Get.create() lifecycle

roipeker:
Replying to an issue spotted by @Quent on Discord, this fixes the lifecycle
( `onInit()`, `onReady()`).
`onClose()` has to be called manually by the dev in the created instance.


example/android/app/src/main/kotlin/com/example/example/MainActivity.kt
```diff
@@ -0,0 +1,6 @@
+package com.example.example
```
Grohden, line 1: duped package name 🤔, did you rebased your branch based on latest master today?
I've fixed a missing activity in the example app

example/android/.gitignore
```diff
@@ -0,0 +1,11 @@
+gradle-wrapper.jar
+/.gradle
+/captures/
+/gradlew
+/gradlew.bat
+/local.properties
```
Grohden, line 6: would be nice to move the top level ignores for android to this file 🤔

example/android/app/src/main/kotlin/com/example/example/MainActivity.kt
roipeker: Probably mate, yep.

Grohden: You can delete it if that's the case

The correct package is `getx.demo.app.example`

But I see that I've created the activity on the java folder instead of the kotlin one.. 🤔

roipeker: @Grohden Do you want me to update the PR with that change?... I'm still new to Github contribution, sometimes it gets a little bit frustrating to me to keep it up with other changes, PRs, etc.

roipeker: What happens with the reference to [java/](https://github.com/jonataslaw/getx/blob/9657bf1d24ce0a996900b848b3b255d05501a8a9/example/android/app/src/main/java/getx/demo/app/example/MainActivity.kt) ? should _MainActivity.kt_ be there?

Grohden: Yup, can be on this PR.

If possible, delete the created activity, the one I've added should work

There's no issue in putting a kotlin file in the 'java' folder, a more 'modern/correct' way is to put them under kotlin folder, but that's only for organisation.. so I don't see a reason to touch the file in this PR.

But try to be sure that the example app still runs after deleting your created activity.

jonataslaw:
Now, LGTM. 
I'll merge
Thanks!

PR #590, opened 2020-09-10
Contribution videos and articles in Readme

stefandevo:
I added a section to summaries all known articles and videos of GetX. 
I moved the existing links and added some new ones I found. 
Section is now at the bottom.

To check if we have to copy to other languages.

I also used https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one plugin in VSCode; it fixed some markdown and automates the TOC.


jonataslaw:
Nice!
LGTM

PR #591, opened 2020-09-10
fix isNullOrBlank function

MrOlolo:
There is no support switching by type in the Dart. You should use a sequence of "if" instead

### DIFF
--- a/lib/src/utils/regex/get_utils.dart
+++ b/lib/src/utils/regex/get_utils.dart
@@ -14,15 +14,10 @@ class GetUtils {
   static bool isNullOrBlank(dynamic s) {
     if (isNull(s)) return true;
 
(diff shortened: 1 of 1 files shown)


Nipodemos (approved):

jonataslaw:
LGTM

Grohden:
@jonataslaw and @Nipodemos the fix is for just only a method.. but there are others with those exactly switches...

@MrOlolo did you had problem with other methods that you could report? I'm gonna open a PR for that if that's the case

MrOlolo:
> @MrOlolo did you had problem with other methods that you could report? I'm gonna open a PR for that if that's the case
I didn't check another methods but they also use type switching which dart not support. So PR is necessary.

P.S. sorry for late answer

PR #592, opened 2020-09-10
fix(num) Fix num issues with operator overloads

Grohden:
As I have commented [here](https://github.com/jonataslaw/getx/pull/565#issuecomment-689053392), num operator overloads are broken, this PR fixes it

There are some situations where it's not possible to overload a operator correclty, for example
the `/` always produces a `double`, so the `RxInt` can't have it.

I would **strongly** recommend get to not overload operators, operators should return new instances completely, get is mutating it's observables

For example:
```dart

final a = RxInt(0);
a + 1;
print(a.value);
```
What's the expected output based on current overloads?

I can't also predict what something like 'a + 1 + 2 + 3 + 4' would emit into it's subjects, (it would probably emit a lot of unnecessary events)

### DIFF
--- a/README.md
+++ b/README.md
@@ -632,10 +632,6 @@ final name = 'GetX'.obs;
 // only "updates" the stream, if the value is different from the current one.
 name.value = 'Hey';
 
-// this weird (and kinda cool) assignment, updates the stream no matter what
-// it takes nulls, or same value... but rebuilds the observers.
-name << 'Hey'; // !
-
 // All Rx properties are "callable" and returns the new value.
 // but this approach does not accepts `null`, the UI will not rebuild.
 name('Hello');
@@ -647,7 +643,7 @@ name() ;
 
 final count = 0.obs;
 
-// you can just most basic operators acts on the property!
+// You can use all non mutable operations from num primitives!
 count + 1;
 
 // Watch out! this is only valid if `count` is not final, but var
@@ -716,10 +712,6 @@ user.update((value){
 });
 
 print( user );
-
-// this also works.
-user << user.value;
-
 ```
 
 #### GetView
--- a/lib/src/state_manager/rx/rx_core/rx_impl.dart
+++ b/lib/src/state_manager/rx/rx_core/rx_impl.dart
@@ -3,6 +3,8 @@ import 'dart:collection';
 
 import '../rx_core/rx_interface.dart';
 
+part 'rx_num.dart';
+
 /// global object that registers against `GetX` and `Obx`, and allows the
 /// reactivity
 /// of those `Widgets` and Rx values.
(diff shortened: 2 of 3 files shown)


Grohden:
Now that I've been thinking better, all operators should only return pure/primitive types:

RxInt(1) + 1  -> expects 2 to be the output, not RxInt(2), if you wish to update the value you should use something like RxInt(0).value = newValue

Grohden:
Oh well, gonna rewrite those operators, anyone that relied on that behavior was wrong... even RxString is wrong

Grohden:
I've amended the changes that properly fix those operators

jonataslaw:
I think that only setter operators (such as "<<") should do subject.add, since it matters in changing the state.
+ maybe, Because of "++"

Grohden:
@jonataslaw those operators can be chained?
`x >> y >> p`? I find weird to chain subject emissions instead of sending them all at once.. but that's may opinion

If you wish I can review all the [overridable operators](https://dart.dev/guides/language/language-tour#overridable-operators) and add the mutable ones

Grohden:
I don't think that >> or << shifts are mutable 🤔 

```dart
void main() {
  final a = -1;
  final b = -1;

  print(a << 2); // 4294967292
  print(b >> 2); // 4294967295
  print(a); // -1
  print(b); // -1
}
```

They don't see to be mutating.. this means that the `_RxImpl<T> operator <<(T val)` is wrong (and also it shouldn't be working for doubles)

Grohden:
@jonataslaw I'm rewritting all operations for nums.

I've stumbled across the RxBool, which is also with wrong operators

```

  bool operator &(bool other) => other && value;

  bool operator |(bool other) => other || value;

  bool operator ^(bool other) => !other == value;
```

Gonna fix them too. This PR will become big tho :/

roipeker:
I don't think anyone is using this, as they were introduced a week ago. And added in docs 2 days ago. Please, remember to update the README/docs if u change the behaviour.

Grohden:
> Please, remember to update the README/docs if u change the behaviour.

🤔 Ok, gonna have a look at docs then.

I've pushed all changes for num/double/int there, I've copied all of them from their original files, Rx* basically becomes a proxy to the _value operations.

IF it's not desirable to have all those operations, I can rever the commit, and/or remove any methods

roipeker:
This came out from the idea of reducing the required code to modify the value primitives. As List,Map,Set were totally reimplemented to match their `.value` signatures... we though of doing (partially) the same for the operators in unsupported native values. I don't know if someone use it, and i'm aware is a little "strange" or confusing for people. So i dont have further opinions about this operator overload feature. Maybe @jonataslaw does...

jonataslaw:
Wow, I really liked that, as it makes the Rx class itself much more concise, without needing to access a ".value" every time you need to access an operator.
For me it's ok,

Grohden:
The only part of the doc I see that has changed is:

```markdown
// this weird (and kinda cool) assignment, updates the stream no matter what
// it takes nulls, or same value... but rebuilds the observers.
name << 'Hey'; // !
```
Which I see that @roipeker introduced... with my changes the shift operator is now present in nums, so I think that a overload with different behaviour is confusing, I understand the intention and it being a shortcut for = is nice, but for anyone reading I would find it confusing in code - but again, that's my opinion.

Nipodemos:
unfortunaly i agree with @Grohden 
we need to remove it

Grohden:
I've only removed refs/changed docs on base readme, I couldn't find corresponding code in translated readmes so.. translations may be outdated :/

If no more action is needed this PR is ready for merge

Nipodemos:
> I couldn't find corresponding code in translated readmes

no one had translated that part yet, so i'ts fine

Nipodemos (approved):
LGTM

@jonataslaw what you think?

jonataslaw:
LGTM, thanks!

PR #593, opened 2020-09-10
Fixes subscription for Rx::bindStream

roipeker:
- add docs to bindStream for RxList,RxSet,RxMap,RxImpl.
- fixes the error of binding an ongoing Stream, without closing the Subscription when the observable gets closed.


jonataslaw:
LGTM

Issue #571, opened 2020-09-06
GetxController onInit and onReady methods called twice


na2axl:
**Describe the bug**
It seems that for a specific case, `onInit` and `onReady` callback are called twice, the first time when the controller is added using `Get.put` (which is normal), and the second time when the controller is used in a `GetX` widget (which I think the initialization must be avoided if already done). This lead for me to create animation controllers and run HTTP queries two times...

**To Reproduce**
1. Create a simple controller
2. Call Get.put() with this controller, assign a tag to it
3. Create a page using this controller, instantiate the controller using a custom tag and then use it in `GetX` widget call, as the `init` parameter
4. The controller is initialized the first time when calling `Get.put`, and the second time in the `initState` of the `GetX` widget

**Expected behavior**
get should know when it is necessary to initialize a controller.

**Flutter Version:**
1.20.0

**Getx Version:**
3.8.0

**Describe on which device you found the bug:**
Samsung SM A7000

**Minimal reproduce code**
```dart
class PageController extends GetxController {
  final RxString _text = "Example".obs;
  String get text => _text.value;
}

class PageBindings extends Bindings {
  final String tag;

  PageBindings({
    @required String tag,
  });

  @override
  void dependencies() => Get.lazyPut(() => PageController(), tag: this.tag);
}

class Page extends StatelessWidget {
  final PageController controller;

  Page({
    Key key,
    @required String tag,
  }) : controller = Get.find(tag: tag),
      super(key: key);

  @override
  Widget build() => Container(
    child: GetX(
      init: controller,
      global: false,
      builder: (_) => Text(controller.text),
    ),
  );
}
```

jonataslaw:
Hi, initially I need to explain what you are doing:

```dart
  @override
  void dependencies() => Get.lazyPut(() => PageController(), tag: this.tag);
```

Here you have created a dependency of type PageController, with a tag.

```dart
  @required String tag,
  }) : controller = Get.find(tag: tag),
```
At that time the controller was created by calling Get.find() (Controller PageController, tag: "your tag" )

```dart
GetX(
      init: controller,
      global: false,
```
Here you used global = false.
This will literally create a new internal instance and pass it on to the children.
When creating an inner class, onInit is called.

```dart
class PageController extends GetxController {
  final RxString _text = "Example".obs;
  String get text => _text.value;
}

class PageBindings extends Bindings {
  final String tag;

  PageBindings({
    @required String tag,
  });

  @override
[...]

PR #595, opened 2020-09-11
[Feature] Ability to use tags with GetX widgets

na2axl:
A simple update to facilitate the use of tagged controllers in a `GetX` widget.

### Before

```dart
class SimplePage extends StatelessWidget {
  final Controller controller = Get.find(tag: "my-tag");

  @override
  Widget build() {
    return GetX(
      // _ is null :-(
      builder: (_) => Center(child: Text(controller.text)),
    );
  }
}

class ComplexPage extends StatelessWidget {
  final Controller controller1 = Get.find(tag: "my-tag-1");
  final Controller controller2 = Get.find(tag: "my-tag-2");

  @override
  Widget build() {
    return Column(
      children: <Widget>[
        Expanded(
          child: GetX(
            // _ is null :-(
            builder: (_) => Center(child: Text(controller1.text)),
          ),
        ),
        Expanded(
          child: GetX(
            // _ is null :-(
            builder: (_) => Center(child: Text(controller2.text)),
          ),
        ),
      ],
    );
  }
}
```

### After

```dart
class SimplePage extends StatelessWidget {
  @override
  Widget build() {
    return GetX<Controller>(
      tag: "my-tag",
      builder: (controller) => Center(child: Text(controller.text)),
    );
  }
}

class ComplexPage extends StatelessWidget {
  @override
  Widget build() {
    return Column(
      children: <Widget>[
        Expanded(
          child: GetX<Controller>(
            tag: "my-tag-1",
            builder: (controller) => Center(child: Text(controller.text)),
          ),
        ),
        Expanded(
          child: GetX<Controller>(
            tag: "my-tag-2",
            builder: (controller) => Center(child: Text(controller.text)),
          ),
        ),
      ],
    );
  }
}
```

### Side notes

I've also fixed a little issue in `GetWidget`, the introduced `tag` feature was not used when the controller is created using `Get.find()`

### DIFF
--- a/lib/src/state_manager/rx/rx_widgets/rx_getx_widget.dart
+++ b/lib/src/state_manager/rx/rx_widgets/rx_getx_widget.dart
@@ -21,7 +21,7 @@ class GetX<T extends DisposableInterface> extends StatefulWidget {
   final String tag;
 
   const GetX({
-    this.tag = null,
+    this.tag,
     this.builder,
     this.global = true,
     this.autoRemove = true,
--- a/lib/src/state_manager/rx/rx_widgets/rx_getx_widget.dart
+++ b/lib/src/state_manager/rx/rx_widgets/rx_getx_widget.dart
@@ -47,22 +47,22 @@ class GetImplXState<T extends DisposableInterface> extends State<GetX<T>> {
   @override
   void initState() {
     _observer = Rx();
-    var isPrepared = GetInstance().isPrepared<T>(tag: tag);
-    var isRegistered = GetInstance().isRegistered<T>(tag: tag);
+    var isPrepared = GetInstance().isPrepared<T>(tag: widget.tag);
+    var isRegistered = GetInstance().isRegistered<T>(tag: widget.tag);
 
     if (widget.global) {
       if (isPrepared) {
         if (GetConfig.smartManagement != SmartManagement.keepFactory) {
           isCreator = true;
         }
-        controller = GetInstance().find<T>(tag: tag);
+        controller = GetInstance().find<T>(tag: widget.tag);
       } else if (isRegistered) {
-        controller = GetInstance().find<T>(tag: tag);
+        controller = GetInstance().find<T>(tag: widget.tag);
         isCreator = false;
       } else {
         controller = widget.init;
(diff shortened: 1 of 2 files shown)


Nipodemos:
very nice addition :D 
but maybe is good to add a check to see if the person is using the init parameter, he/she can't use the tag parameter

and what are the implications with the `global: false` parameter? the little I understand the controller will be create just for the widgets inside the GetX widget, so in this case we can't use tag too right?

Nipodemos:
i think we should put a Exception to prevent users to use the tag together with init

na2axl:
Yes @Nipodemos, I've tested all theses case before falling on adding the `tag` field on the `GetX` widget, I've even created an issue #571 

Without the `tag` field, if the user want to use a tagged controller with `GetX` and use the `init` field like `init: Get.find<Controller>(tag: "my-tag")` this will work, but a little problem will be that since `global` is `true` on this case, Get will `put` this controller in the dependency system **without a tag**, and this is not the good behavior on that case for me...

However, if we try to avoid it by using `global: false` and `init: Get.find<Controller>(tag: "my-tag")`, this will lead to the issue #571, where `onInit` and `onReady` are called twice, the first time implicitly by `Get.find()`, and the second time explicitly by `GetX`.

To avoid these two conflicts in behaviors I think it's better to let `GetX` `find` the right tagged controller himself.

So yes, when using the `tag` field, the user must not use `init` again, otherwise if he want a kind of _fallback_ if the given tag doesn't exists in the dependency system at that time, this way, if `init` and `tag` are provided, and `GetX` doesn't found any controller with the given tag, it will `put` the controller provided by `init` **with the right tag** (which is the current behavior without using a tag).

Nipodemos:
wait, so what @jonataslaw suggested in #571 was non-existent until this PR? 🤔 

well, this PR sounds good enough for me 😄

but i'm not the one that can check code
let's wait for jonatas to see his opinion

jonataslaw:
I'm sure I included "tag" in GetBuilder and GetX in some PR, I don't understand how it was lost. I must have done some regression.
So thanks for taking this.
LGTM

Issue #600, opened 2020-09-12
RTL directionality support


smokelaboratory:
Wonderful package!
Is there a way to change layout direction for RTL languages when using GetX translations?

roipeker:
Thanks for noticing that @smokelaboratory ! 
I didn't try the internationalisation capabilities of GetX™ until now.
I just pushed a fix to refresh the UI when you `Get.changeLocale()`, and included in the docs and [README](https://github.com/jonataslaw/getx/blob/7754903adb000ae3712ccad4e07b4922ef323c30/README.md#locales).
Please, give it a try when it gets merged, and let me know if the issue is solved.

Hamidrezana:
@roipeker 

I have same issue, and I try what you included in README, but still have issue.
layout does not update on changing locale, but update after after hot reload.

```
  Widget build(BuildContext context) {
    return StoreProvider<AppState>(
      store: store,
      child: AppThemeProvider(
        child: (themeContext) => GetMaterialApp(
          title: Strings.appName,
          debugShowCheckedModeBanner: false,
          localizationsDelegates: [
            GlobalMaterialLocalizations.delegate,
            GlobalWidgetsLocalizations.delegate,
            GlobalCupertinoLocalizations.delegate,
          ],
          translations: Translation(),
          locale: Locale('fa', 'IR'),
          fallbackLocale: Locale('en', 'US'),
          supportedLocales: [Locale('fa', 'IR'), Locale('en', 'US')],
          popGesture: true,
          theme: ThemeProvider.themeOf(themeContext).data,
          home: InitialApp(),
        ),
      ),
    );
  }
```

smokelaboratory:
@Hamidrezana 

I think the change is not yet merged. It is under review.
Here is the PR for it (https://github.com/jonataslaw/getx/pull/601)

Hamidrezana:
@smokelaboratory tnx.

jonataslaw:
closed by https://github.com/jonataslaw/getx/commit/ccf017823a0b3bb41bccec11653d54dc29015579

roipeker:
wonderful approach

virodveloper:
Hy  here is the temporary fix To Change Language using GetX utils(RTL and LTR)
Step 1 : import flutter_localizations SDK
Step 2 : Make your MyApp base Widget to stateful
Step 3 : Write these lines in MyApp base widget
static void setLocale(BuildContext context,Locale locale){
[...]

PR #601, opened 2020-09-12
Localization updates

roipeker:
- fixes issue #600
- moved (partially) Get.locale to rootController (should be in it's own package domain)
- fixed `updateLocale` to invalidate `MaterialApp`
- added `Get.systemLocale` and `Get.systemLocale` to suggestion of using `dart:ui`
- updated comment docs in several places.
- updated README with internalization details and steps to include  flutter_localizations (which should be mandatory).


Nipodemos (approved):

Nipodemos:
sounds nice to me 😄 
now we have the basics of how to make your app accept localizations, but do we have docs of how to use getx localization?

roipeker:
I push the update to the README.

Nipodemos (approved):
Sounds good to me 😄

README.md
```diff
@@ -313,18 +313,38 @@ class Messages extends Translations {
...
-Just append `.tr` to the specified key and it will be translated, using the current value of `Get.locale` and `Get.fallbackLocale`.
+Just append `.tr` to the specified key, and it will be translated, using the current value of `Get.locale` or `Get.fallbackLocale`.
 
 ```dart
 Text('title'.tr);
 ```
 
 ### Locales
 
+Flutter, by default, only supports English language `Locale` (*en_XX*).
```
jonataslaw, line 324: Just to be more precise, could you tell us in what situations this occurs?
When I use the GetMaterialApp locale in PT_br, I have no exceptions

lib/src/navigation/extension_navigation.dart
```diff
@@ -903,8 +905,13 @@ extension GetNavigation on GetInterface {
     }
   }
 
+  /// Changes the current Locale of your app.
+  /// Make sure you import [flutter_localizations] in your pubspec.yaml,
+  /// and add the [localizationDelegates] as described:
+  /// https://flutter.dev/docs/development/accessibility-and-localization/internationalization
+  /// Otherwise, you will get an Exception.
   void updateLocale(Locale l) {
```
jonataslaw, line 913: Couldn't that possibly cause a double rebuild?

jonataslaw:
Thanks for that PR, it's a really good fit in the docs, and in the code, and Get.defaultLocale is cool.
I just have a few caveats:

1- In web projects, using the flutter_localizations sdk is often not an option. In my last project this was vetoed due to the significant increase in the .js file.
I think maybe it would be better to demonstrate how to use it without any package (like now), and how to use it in conjunction with flutter_localizations.

2- I am afraid that using update + forceAppUpdate will cause double rebuilds. I haven't tested it, but looking at the code, it looks like it does it.

roipeker:
Regarding
1 - Check [this article](https://medium.com/flutter/optimizing-performance-in-flutter-web-apps-with-tree-shaking-and-deferred-loading-535fbe3cd674) for Deferred loading localizations. So, at least in my experience, if you changed the Locale, and dont update the MaterialApp, and then you change the ThemeMode, app will crash cause lots of Material Widgets have localized Strings (like semantics or toolrips) that can't be found. If there's a workaround for that, please let me know so I can update the README... but according to my tests, `flutter_localizations` was a forced things to add.

2 - I believe you can't escape the `forceAppUpdate()` call, and certainly need to update MaterialApp to apply the update to material widgets (RTL change for instance), several users had the issue of not having the app refreshed, forcing a call to `Get.changeTheme`/`Get.changeThemeMode` to invalidate the rootController.

README.md
roipeker: Idk why it happens to me.

jonataslaw:
> Thanks Jonny:
> 
> 1 - Check [this article](https://medium.com/flutter/optimizing-performance-in-flutter-web-apps-with-tree-shaking-and-deferred-loading-535fbe3cd674) for Deferred loading localizations. So, at least in my experience, if you changed the Locale, and dont update the MaterialApp, and then you change the ThemeMode, app will crash cause lots of Material Widgets have localized Strings (like semantics or tooltips) that can't be found. If there's a workaround for that, please let me know so I can update the README... but according to my tests, `flutter_localizations` was a forced things to add.

Deferred loading does not reduce the size of the .js file in Flutter (unfortunately).
Material is not used in web applications due to the size of the .js file (at least in my projects, it is always necessary to comment on the "use-material: true" of pubspec, and all material widgets that use icons, such as datepicker are redone from scratch).
I created internationalization for a real need in a project that I was inserted to have less than 800kb of site load, and I believe this is a recurring problem in Flutter, so the localization sdk, in my opinion, cannot be mandatory . It may even be suggested, but not mandatory, precisely because of the Flutter web file size. Well, at least in my experience with Flutter web in professional projects it required this.

> 2 - I believe you can't escape the `forceAppUpdate()` call, and certainly need to update MaterialApp to apply the update to material widgets (RTL change for instance), several users had the issue of not having the app refreshed, forcing a call to `Get.changeTheme`/`Get.changeThemeMode` to invalidate the rootController.

You are right, but I will try to think of an alternative to improve this in the future.


Thank you for another contribution.
I would just like to see the change in the sdk and internationalization part in a separate topic, such as "using get with flutter_localizations", and inform that this is necessary for RTL and for internationalization of Flutter's internal widgets.

roipeker:
@jonataslaw , sure i will change that in the README. You sure you have no exceptions when passing a different Locale?

jonataslaw:
> @jonataslaw , sure i will change that in the README. You sure you have no exceptions when passing a different Locale?

Yep, as we talked about yesterday, the exception is only thrown when "supportedLocales" is filled.
In addition, I did a PR to support RTL, so the use of the Flutter_internacionalizations SDK is restricted to who will use Material. As Material is not used on the web due to the size of the js file, I believe that the SDK part should be in a separate topic for those who want to use both

README.md
eduardoflorence: I think what should be changed in the documentation is just the removal of this part:
```dart
  ...
  supportedLocales: <Locale> [Locale ('en', 'UK'), Locale ('en', 'US'), Locale ('de', 'DE')]
  ...
```
Because it is not necessary for GetX's Internationalization feature and it still causes an error that needs to be resolved with flutter_localizations.

supportedLocales is useful for Material components, for example, for the calendar component to show the names of months in the defined language

jonataslaw:
This was a good PR, I will reuse it in the current version

README.md
Nitingadhiya: @eduardoflorence 
I used GetX package in my app. now try to implement Calendar with locale but it's give me some error.
Error:- "No MaterialLocalizations Found. DatePickerDialog widget require MaterialLocalization to be provided by a Localization widget ancestor".

Please support me how i can fix that error?

PR #603, opened 2020-09-12
change `arguments` to dynamic (roadmap)

### LINKED ISSUE #441: In Get.arguments change object to dynamic
**Fill in the template. Issues that do not respect the model will be closed.**

**Describe the bug**
When i input map to `Get.toNamed` arguments, thats not working to call that map.

```dart
print(Get.arguments['mapKey']);
```

Will throw error like this

```
The operator '[]' isn't defined for the type 'Object'. Try defining the operator '[]'.
```

**Expected behavior**
Working like if in type of dynamic. If i convert to dynamic, working as i expected.

```dart
dynamic arguments = Get.arguments;
print(arguments['mapKey']); // This working
```

**Flutter Version:**
1.17.4

**Get Version:**
3.4.1

My english is very bad, i'm sorry.

roipeker:
- Fixes #441
- I understand `Object` was used, because Flutter's Route arguments uses it. Yet, seems it will not bring any breaking change, beyond removing all type checking to the `arguments` property. So, it gives full freedom (and responsibility) to the developer to make the proper validations.

Check responses in (we might need to validate the usage of extension methods over `dynamic`):
https://stackoverflow.com/questions/31257735/what-is-the-difference-between-dynamic-and-object-in-dart#:~:text=Writing%20dynamic%20o%3B%20declares%20a,except%20that%20it%27s%20an%20Object%20.

### DIFF
--- a/lib/src/navigation/extension_navigation.dart
+++ b/lib/src/navigation/extension_navigation.dart
@@ -47,7 +47,7 @@ extension GetNavigation on GetInterface {
     Duration duration,
     int id,
     bool fullscreenDialog = false,
-    Object arguments,
+    dynamic arguments,
     Bindings binding,
     bool preventDuplicates = true,
     bool popGesture,
@@ -93,7 +93,7 @@ extension GetNavigation on GetInterface {
   /// Note: Always put a slash on the route ('/page1'), to avoid unnexpected errors
   Future<T> toNamed<T>(
     String page, {
-    Object arguments,
+    dynamic arguments,
     int id,
     bool preventDuplicates = true,
   }) {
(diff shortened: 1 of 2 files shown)


Nipodemos:
if i do remember correctly, i think a long ago it was Objects, but it was giving problems when the argument was a list or a Map
although it was so long ago that i'm not sure if i'm right

@jonataslaw do you remember something like that?

jonataslaw:
In the past it was dynamic, but for the sake of standardization I switched to Object, because Flutter uses it as Object.
Well, I have nothing against this change at the moment, and it can allow more flexibility, so LGTM.
Thank you Roi!

Issue #503, opened 2020-08-25
[Feature Request]: Adding Persistent Bottom Sheet.


Nipodemos:
@GoldenSoju suggested on [a comment in issue #201 ](https://github.com/jonataslaw/getx/issues/201#issuecomment-662974044)

One favor I would like to ask for:

### Adding Persistent Bottom Sheet.

At the moment "Get.bottomSheet()" returns a Modal Bottom Sheet, but I was wondering if you could also add the Persistent Bottom Sheet ?

**Modal:** https://api.flutter.dev/flutter/material/showModalBottomSheet.html
**Persistent:** https://api.flutter.dev/flutter/material/showBottomSheet.html

Modal Bottom Sheet makes the rest of the screen inactive but
I would need the behavior of the Persistent Bottom Sheet where the user can interact with the rest of the screen when the bottom sheet is shown. 

Obs: i am separating feature request in different issues so they can be put in the projects board
@jonataslaw if you don't plan to make this feature, tell me so this issue can be closed and removed from the roadmap
this can be checked [here](https://github.com/jonataslaw/getx/projects/1?add_cards_query=is%3Aopen)

jonataslaw:
It is an easy feature to implement, it will be implemented in the short term, after this roadmap:

https://github.com/jonataslaw/get_cli

jonataslaw:
Adding to projects sounds incredible to me, and it will become more organized than a simple roadmap. Thank you.

mohak852:
@jonataslaw I wanna contribute to this getx should i try to solve this issue

Nipodemos:
@mohak852 every contribution is very welcomed ❤️

roipeker:
> @jonataslaw I wanna contribute to this getx should i try to solve this issue

Any progress?

jonataslaw:
Added https://github.com/jonataslaw/getx/pull/604

PR #604, opened 2020-09-12
Added Persistent bottom sheet 

mohak852:
#503 Added Persistent bottom sheet in getx.


roipeker:
@mohak852 How do you use the Bottomsheet? 
`_GetPerModalBottomSheet ` is a private class, with no external API reference, am I missing something?

Nipodemos:
Hello!
I come here just to ask for you to not send the pubspec.lock file.

In the end @jonataslaw will overwrite the file with his own when he pushes his commits, so there is no point in doing that
the rest of the review i'll leave for jonatas to see 😅

mohak852:
@roipeker Actually currently I am working on the final update I have just send the pr so that you can view the __GetPerModalBottomSheet_ class. If any changes in this, please refer to me.

mohak852:
I have a doubt that how we are calling **_Get.anycoomand_** in the dart files.

mohak852:
Waiting for your reply.

roipeker:
the call should be at the same [level/file](https://github.com/jonataslaw/getx/blob/master/lib/src/navigation/extension_navigation.dart) as `Get.bottomSheet`.
Regarding the name, I have no idea, if it's the same method signature for **modal** and **persistent** bottomSheets, except that persistent happens at Scaffold's level (should have the option to not block the UI underneath it)... I would just add a "persistent: true/false" in the bottomSheet call. Otherwise maybe use `rawBottomSheet()` or `persistentBottomSheet()` ... anyway, if it works, and gets approved, we have a few days to change the name, before next release.

mohak852:
@roipeker as suggested by you I have made the persistent bottom sheet as bool and now you can check this.

jonataslaw:
LGTM, thanks for yout contribution!

PR #607, opened 2020-09-13
Add more tests for string and num extensions

Nipodemos:
Hello!

As I noticed an increasing amount of tiny bugs that could've been detected with unit tests, I decided to write more tests.

But i never wrote tests for anything in my entire life. So I started with simple things to get the hang of it (also because I don't understand GetX source code enough to test it).

I went for number extensions, and after String, and it took me much longer then expeced to write all tests just for string extensions. So I did only that. I want to do more in the future.
But in the end it was a good thing, because I found several errors and fixed some of them.

### What was fixed:
* camelCase was considering only the first two words
* isBinary was returning true even for empty strings
* isPalindom wasn't working with phrases
* isCurrency wasn't accepting `R$10.00`
* capitalized was not capitalizing correctly
* there was a duplicated file of the extensions, probably lost itself in the middle of merge conflicts
* commented out print statements in tests

### BREAKING CHANGES:
* `string.isCpf(string)` now is a getter, needs to be used like this: `string.isCpf`
* `sring.isCnpj(string)` now is a getter, needs to be used like this: `string.isCnpj`
* `string.capitalize(string)` now is a getter, needs to be used like this: `string.capitalize`. Also, it does not have anymore the `firstOnly` paramenter
* `string.capitalizeFirst (string)` now is a getter, needs to be used like this: `string.capitalizeFirst `
* `string.removeAllWhitespace (string)` now is a getter, needs to be used like this: `string.removeAllWhitespace `
* `string.camelCase (string)` now is a getter, needs to be used like this: `string.camelCase `
* `string.numericOnly(string, firstWordOnly = false)` now does not have the string parameter, only the optional firstWordOnly
#### Note:
If you guys noticed, the problem with that methods is that they are asking a string as parameter, but they should not, since it is a extension of a string

### What does not work:
* I could not discover how to make dynamic extensions to work. Can anyone show me an example of how to use `var.isNullOrBlank` so i can make tests of it?
* Several ways of typing a phone number are not working, for example:
`(31) 9 9999-9999` 
`(64) 99999-9999`
`+1 202-555-0145`
It is incredibly hard to support all phone numbers. The way it is now, it not not reliable for real world app that are international.
* isPalimdrons does not support UTF8 characters, because regex does not support. Only english letters. I tried to search a little of how to make it work but i found hard for me to do it.
* `isCurrency` only supports like 20 different currencies, althought it will cover most of uses, it does not cover all. If cover all currencies in a regex is relevant enough I don't know, myabe it isn't.

Well i guess this is all for today. This PR is ready to merge after you review it @jonataslaw


stefandevo:
Nice work. I do think that some of these extensions are not complete enough to be in the package. I agree with @Nipodemos that phone number, currency etc are not complete and therefore should not be in the package. I think there are several other packages that have a 100% focus on this.

Nipodemos:
Updated dynamic extensions, now I realized what mistake was I making.
Everything working great

@jonataslaw feel free to merge if you want

jonataslaw:
Very Nice! 
Thanks for that. 
I'll merge now

PR #623, opened 2020-09-16
Fix for switch types usages in GetUtils

Grohden:
This PR fixes the missing functions that were using 'switch types' (I've made a comment in #591), I've also **deprecated** the `isLengthLowerThan` and `isLengthLowerOrEqual` functions, I've **renamed** them to `isLengthLessThan` and `isLengthLessOrEqual` respectively.

I've also added some tests, I would add tests for all fns but there's a lot to test 😭...  perhaps they can be added by need..

### DIFF
--- a/test/src/utils/get_utils_test.dart
+++ b/test/src/utils/get_utils_test.dart
@@ -1,0 +1,56 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:get/get.dart';
+
+// Identity util to convert to iterables
+dynamic _id(dynamic e) => e;
+
+void main() {
+  // Tests for GetUtils.isNullOrBlank are located at dynamic extensions
+  group('GetUtils.isLength* functions', () {
+    test('isLengthEqualTo should validate iterable lengths', () {
+      // iterables should cover list and set
+      expect(GetUtils.isLengthEqualTo([].map(_id), 0), true);
+      expect(GetUtils.isLengthEqualTo([1, 2].map(_id), 2), true);
+
+      expect(GetUtils.isLengthEqualTo({}, 0), true);
+      expect(GetUtils.isLengthEqualTo({1: 1, 2: 1}, 2), true);
+      expect(GetUtils.isLengthEqualTo({}, 2), false);
+
+      expect(GetUtils.isLengthEqualTo("", 0), true);
+      expect(GetUtils.isLengthEqualTo("a", 0), false);
+      expect(GetUtils.isLengthEqualTo("a", 1), true);
+    });
+
+    test('isLengthGreaterOrEqual should validate lengths', () {
+      // iterables should cover list and set
+      expect(GetUtils.isLengthGreaterOrEqual([].map(_id), 0), true);
+      expect(GetUtils.isLengthGreaterOrEqual([1, 2].map(_id), 2), true);
+      expect(GetUtils.isLengthGreaterOrEqual([1, 2].map(_id), 1), true);
+
+      expect(GetUtils.isLengthGreaterOrEqual({}, 0), true);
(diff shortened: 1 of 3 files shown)


test/src/utils/get_utils_test.dart
```diff
@@ -0,0 +1,97 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:get/get.dart';
+
+// Identity util to convert to iterables
+dynamic _id(dynamic e) => e;
+
+void main() {
+  group('GetUtils.isNullOrBlank', () {
```
Nipodemos, line 8: so... about these tests

What do you thing of putting them on dynamic extensions?
I've done tests for all methods that are used as extension for String, num, and dynamic. It covered some of your tests, but it did not covered all of them.

So it would be awesome if you merge them
There are a few tests of isNullOrBlank there, what you think of put them there?

Nipodemos (changes requested):

test/src/utils/get_utils_test.dart
Grohden: okay, gonna merge them with those on dynamic exts

Grohden:
Just my 2 cents here for another PR:
I would deprecate the `isNullOrBlank` name and rename it to `isNullOrEmpty` instead... because that's what the `isNullOrBlank` is currently doing - it also trim compare strings, but that **was not working correctly** before this PR, so renaming and removing the trim compare would preserve the **current** behaviour of isNullOrEmpty.

For me the `isNullOrBlank` only make sense for strings

test/src/utils/get_utils_test.dart
Grohden: @Nipodemos can you resolve this conversation once you've confirmed my test merges with those on dynamic are ok?

jonataslaw:
> Just my 2 cents here for another PR:
> I would deprecate the `isNullOrBlank` name and rename it to `isNullOrEmpty` instead... because that's what the `isNullOrBlank` is currently doing - it also trim compare strings, but that **was not working correctly** before this PR, so renaming and removing the trim compare would preserve the **current** behaviour of isNullOrEmpty.
> 
> For me the `isNullOrBlank` only make sense for strings

I actually wanted to do this for the longest time.
At the time I did GetUtils, I just came out creating methods and utilities and didn't think about the name. 1 Day after launch I had already regretted the "isNullOrBlank", because in fact, Strings and Iterables can be "null" or "empty", while all other classes can be just null.
I agree with the change, and I believe that another method could be created, and the current one deprecated (to avoid version bump), and in version 4.0 we can remove it.

Grohden:
@jonataslaw I will make another PR deprecating/renaming `isNullOrBlank` method so we can discuss this further

jonataslaw:
LGTM

test/src/utils/get_utils_test.dart
```diff
@@ -0,0 +1,56 @@
...
+
+// Identity util to convert to iterables
+dynamic _id(dynamic e) => e;
+
+void main() {
+  // Tests for GetUtils.isNullOrBlank are located at dynamic extensions
+  group('GetUtils.isLength* functions', () {
+    test('isLengthEqualTo should validate iterable lengths', () {
+      // iterables should cover list and set
+      expect(GetUtils.isLengthEqualTo([].map(_id), 0), true);
```
Nipodemos, line 12: hey, just for me to understand, why have you used map here?
using simply a `[]` will not work?

Nipodemos (approved):
These tests are great ❤️

Nipodemos:
since jonatas already approved, i'll merge

test/src/utils/get_utils_test.dart
Grohden: I wanted to test Iterables, to cover lists and sets.
I couldn't find a better way to make a iterable than using map

PR #627, opened 2020-09-17
RxList null check in the constructor

Hitsu91:
If RxList is instantiated with the constructor without provide the optional parameter:
`RxList([List<E> initial])`
Like:
`RxList<Foo> foo = RxList();`
the constructor will override the property "_list" with a null.
This can cause null pointer exception on methods such as
isEmpty
isNotEmpty
An easy fix is to pass an empty list to the constructor:
`RxList<Foo> rxList = RxList([]);`
But that isn't a good solution, the linter will not flat the use of the constructor without the optional value.


jonataslaw:
Hi, thanks for that!
Can you add it to Map and Set too?

Hitsu91:
> Hi, thanks for that!
> Can you add it to Map and Set too?

Sure

jonataslaw:
Thank's, LGTM

PR #636, opened 2020-09-21
fix(unknownRoute): Fix customTransition being called on null

marcosfons:
CustomTransition being called on null when opening unknown route


jonataslaw:
Thanks for that.
LGTM!

PR #638, opened 2020-09-22
Fix #624

### LINKED ISSUE #624: GetUtils::capitalize method!
```
static String capitalize(String s, {bool firstOnly = false}) {
    if (isNullOrBlank(s)) return null;
    if (firstOnly) return capitalizeFirst(s);

    var lst = s.split(' ');
    var newStr = '';
    for (var s in lst) {
      newStr += capitalizeFirst(s);
    }
    return newStr;
  }
```

This method omits spaces between words once capitalized.
I think below function block is right one.

```
static String capitalize(String s, {bool firstOnly = false}) {
  if (isNullOrBlank(s)) return null;
  if (firstOnly) return capitalizeFirst(s);
  return s.split(' ').map(capitalizeFirst).join(' ');
}
```
 
Thank you.

zl910627:
closes #624

### DIFF
--- a/packages/get_utils/lib/src/get_utils/get_utils.dart
+++ b/packages/get_utils/lib/src/get_utils/get_utils.dart
@@ -506,19 +506,8 @@ class GetUtils {
   /// Capitalize each word inside string
   /// Example: your name => Your Name, your name => Your name
   static String capitalize(String value) {
-    if (isNullOrBlank(value)) {
-      return null;
-    }
-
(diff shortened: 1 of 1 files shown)


Nipodemos (approved):
Tests passed, LGTM!
Now is just wait for @jonataslaw approval

jonataslaw:
Thanks for your contribution
LGTM, I'll merge

PR #647, opened 2020-09-24
Fixes Get.create() and GetWidget()

roipeker:
- Although "nasty" solution, multiple instances were created with one Get.find() when using Get.create(), now the actual instance created from `GetInstance::<S>_startController()` > `GetInstance::<S>_initDependencies()` > `GetInstance::<S>_find()` is returned.
- added docs and impl to `GetWidget` to make it work with **DI** `tag`, I place it in the constructor directly (to use with `:super( tag: "something")` )... but we might actually use an override getter as well if needed.

### DIFF
--- a/packages/get_state_manager/lib/src/simple/get_view.dart
+++ b/packages/get_state_manager/lib/src/simple/get_view.dart
@@ -11,9 +11,13 @@ import 'package:get_instance/get_instance.dart';
 /// }
 ///
 /// class AwesomeView extends GetView<AwesomeController> {
-///   /// if you need you can pass the tag for your controller
-///   /// in the super constructor.
-///   AwesomeView({Key key}):super(key:key, tag:"myControllerTag");
+///   /// if you need you can pass the tag for
+///   /// Get.find<AwesomeController>(tag:"myTag");
+///   @override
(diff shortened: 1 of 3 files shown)


packages/get_state_manager/lib/src/simple/get_view.dart
```diff
@@ -21,9 +24,9 @@ import 'package:get_instance/get_instance.dart';
 /// }
 ///``
 abstract class GetView<T> extends StatelessWidget {
-  const GetView({Key key}) : super(key: key);
+  const GetView({Key key, this.tag}) : super(key: key);
 
-  final String tag = null;
+  final String tag;
```
jonataslaw, line 29: This requires the implementation of the tag, since it is an abstract class. It is strange that it receives null, but unfortunately dart is still confused between implementation and extends. This should not happen, but as it does, it should be maintained

jonataslaw (changes requested):

packages/get_state_manager/lib/src/simple/get_view.dart
roipeker: done....

jonataslaw:
LGTM now
Thanks

PR #663, opened 2020-09-29
typedef over dynamic Function for GetX/GetBuilder builders

jamie1192:
- both GetX and GetBuilder
- now typesafe return instead of empty Function

Now allows for Dart to automatically autofill the required parameter value, whereas before it returned a dynamic Function with no parameter.
Code 2020-09-29 at 13 40 27
After:
Code 2020-09-29 at 13 40 45


Nipodemos (approved):
fancy 😄

jonataslaw:
Thanks!
LGTM!

PR #683, opened 2020-10-06
Fix null route error

FiercestT:
This was using flutter web.
I encountered an error in which I was redirected back to my site, I had 2 issues (1 I could not solve, will make an issue),
the other, caused an error in which `match.route` was null on `initialRoutesGenerate`, and members were being accessed on a null field. This would show the default browser 404 page, even though an unknown route was setup.

This will show the unknown route in that case.

I have also updated the navigation doc to let new users know to setup an unknown route.

### DIFF
--- a/documentation/en_US/route_management.md
+++ b/documentation/en_US/route_management.md
@@ -152,6 +152,23 @@ void main() {
 }
 ```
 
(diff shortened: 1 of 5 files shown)


jonataslaw:
Thanks for that!
LGTM

PR #718, opened 2020-10-17
Refactor Obx and ObxValue with ObxWidget

kranfix:
The `Obx` and `ObxValue` widgets share almost all the logic, then that logic could be modularized with an `ObxWidget` base (worrking like a StatelessWidget or StatefulWidget).  
  
There are no breaking changes. This is only a refactor to reuse code.


jonataslaw:
Hi, this seems much clearer to me. Thank you for that!
LGTM

PR #727, opened 2020-10-20
Navigation Generic && Arguments

unacorbatanegra:
Hello, I just added some generic parameters and change the `RouteSettings`  in the `Get.dialog` to `Object arguments`

### DIFF
--- a/lib/get_navigation/src/extension_navigation.dart
+++ b/lib/get_navigation/src/extension_navigation.dart
@@ -220,7 +220,7 @@ extension ExtensionDialog on GetInterface {
     Color barrierColor,
     bool useSafeArea = true,
     bool useRootNavigator = true,
-    RouteSettings routeSettings,
+    Object arguments,
     Duration transitionDuration,
     Curve transitionCurve,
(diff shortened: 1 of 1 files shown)


lib/get_navigation/src/extension_navigation.dart
```diff
@@ -220,7 +220,7 @@ extension ExtensionDialog on GetInterface {
     Color barrierColor,
     bool useSafeArea = true,
     bool useRootNavigator = true,
-    RouteSettings routeSettings,
+    Object arguments,
```
jonataslaw, line 223: Unfortunately this cannot be changed, first because it would be a break update, and second that it is possible to define Dialog names via routeSettings, something that is not very common, but it is a feature.

jonataslaw (changes requested):

jonataslaw:
Nice!
Thanks for that, I just made a note about the arguments that cannot be exchanged

jonataslaw:
Thanks for your contribution!
LGTM

PR #766, opened 2020-11-05
Fix incorrect Get.reference when pop route

4mb1t:
When pop route, the newly active route is previousRoute.
Ref: [NavigatorObserver.didPop](https://api.flutter.dev/flutter/widgets/NavigatorObserver/didPop.html)


jonataslaw:
Hi, thanks for that!
It makes perfect sense.
LGTM

Issue #804, opened 2020-11-19
Error: Type 'ui.Window' not found.


asimbo:
**ATTENTION: DO NOT USE THIS FIELD TO ASK SUPPORT QUESTIONS. USE THE PLATFORM CHANNELS FOR THIS. THIS SPACE IS DEDICATED ONLY FOR BUGS DESCRIPTION.**
Error: Type 'ui.Window' not found. on start up

**Describe the bug**
ui.Window get window => ui.window;
  ^^^^^^^^^

**Reproduction code
NOTE: THIS IS MANDATORY, IF YOUR ISSUE DOES NOT CONTAIN IT, IT WILL BE CLOSED PRELIMINARY)**
/C:/flutter/.pub-cache/hosted/pub.dartlang.org/get-3.17.1/lib/get_navigation/src/extension_navigation.dart:1047:3: Error: Type 'ui.Window' not found.

FAILURE: Build failed with an exception.

Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel master, 1.24.0-8.0.pre.291, on Microsoft Windows [Version 10.0.18362.1016])
[√] Android toolchain - develop for Android devices (Android SDK version 30.0.0)
[√] Chrome - develop for the web
[√] Android Studio (version 4.0)
[√] VS Code (version 1.51.1)
[√] Connected device (4 available)

imad-dusky:
Same Problem

Akiat:
If you need (like me), to stay on the flutter master branch, the PR #805 fix the issue.

jonataslaw:
add on your pubspec on dependencies:

```yaml
dependencies:
  get:
    git:
      url: git@github.com:jonataslaw/getx.git
      ref: dev
```

Akiat:
Great, thank you 👍 

I used this synthax to avoid a pub get error:

```yaml
  get:
    git:
      url: https://github.com/jonataslaw/getx.git
      ref: dev
```

kyawsithu:
Because get_storage >=1.3.2 depends on get from hosted and ProjectName depends on get from git, get_storage >=1.3.2 is forbidden.

How to solve this?

eduardoflorence:
@kyawsithu use flutter stable channel and GetX 3.17.1

kyawsithu:
Thanks @eduardoflorence issue was solved

jonataslaw:
Fixed on last update

sahoo4784:
1.Restart the IDE
[...]

PR #805, opened 2020-11-19
Possible fix for issue #804

worldofsites:
Possible fix for issue https://github.com/jonataslaw/getx/issues/804


jonataslaw:
This is the solution for the future release. The Api Window was deprecated, and changed to SingletonFlutterWindow on master, however, this PR will break Stable.
So I will wait until this version becomes stable to accept this.

Akiat:
Perfectly worked for me, thanks 👍

Issue #815, opened 2020-11-23
Get.offUntil() arguments


vbuberen:
While working with navigation using GetX I found something that looks inconsistent for me, but I might not know some aspects, so would like to clarify.

Why it is possible to pass widget with other navigation extensions, like `Get.to()`,  `Get.off()`, etc., but not with `Get.offUntil()`, which requires `Route<T> page` as a destination parameter? 
https://github.com/jonataslaw/getx/blob/a27d206d7634e9811b36d4715e37f773b507a97b/lib/get_navigation/src/extension_navigation.dart#L603-L607

Are there any reasons to use exactly this type in parameter in this case? I believe it would be more consistent if users could just pass widget and `Route` will be created by GetX, like it is done in `Get.to()`, for example.

If it is something that was just missed somewhere I would like to try and create a PR with this extension update.

jonataslaw:
> While working with navigation using GetX I found something that looks inconsistent for me, but I might not know some aspects, so would like to clarify.
> 
> Why it is possible to pass widget with other navigation extensions, like `Get.to()`, `Get.off()`, etc., but not with `Get.offUntil()`, which requires `Route<T> page` as a destination parameter?
> https://github.com/jonataslaw/getx/blob/a27d206d7634e9811b36d4715e37f773b507a97b/lib/get_navigation/src/extension_navigation.dart#L603-L607
> 
> Are there any reasons to use exactly this type in parameter in this case? I believe it would be more consistent if users could just pass widget and `Route` will be created by GetX, like it is done in `Get.to()`, for example.
> 
> If it is something that was just missed somewhere I would like to try and create a PR with this extension update.

To be honest, I would like only widgets to be sent, and this is not the only method that has this kind of inconsistency.
I they were inserted somewhere between Get 1.0 and Get 2.0, and only the most popular ones were migrated to widgets, but I believe that this is something that deserves to be improved.
Feel free to do a PR, after I finish the feature I'm working on for the new big update, I will totally turn to improve navigation, including a widget for nestedNavigators.

PR #817, opened 2020-11-24
Update Get.offUntil() extension

vbuberen:
This PR addresses #815 

Get.offUntil() now accepts `Widget` as a destination instead of `Route<T>` to be similar to other navigation extensions for non-named routes.

Also, a set of available optional parameters is the same as in `Get.to()` for more consistency.

### DIFF
--- a/lib/get_navigation/src/extension_navigation.dart
+++ b/lib/get_navigation/src/extension_navigation.dart
@@ -2,6 +2,7 @@ import 'dart:ui' as ui;
 
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
+
 import '../../get_core/get_core.dart';
 import '../../get_instance/src/bindings_interface.dart';
 import '../../get_utils/get_utils.dart';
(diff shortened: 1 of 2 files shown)


lib/get_navigation/src/extension_navigation.dart
```diff
@@ -600,10 +601,44 @@ extension GetNavigation on GetInterface {
   /// or also like this:
```
vbuberen, line 601: I couldn't put the comment on the right line (598), so putting quite lower.

Would like to find out if this comment is outdated, since I can't make the same predicate work for `offUntil()` extensions. After debugging I found out that for this extension it makes no sense to check `currentRoute`, since `currentRoute` will always be the page where user is till the whole logic finishes.

eduardoflorence:
This PR will cause a breaking change, so I think it can only be approved for version 4.x.x of GetX, so that developers using `Get.offUntil()` can be communicated.

vbuberen:
Right, my bad.

I suggest to have 2 functions: the old one and the new one, but add a proper deprecation annotation to the old one.

vbuberen:
However, also not a good idea. I am relatively new to Dart, so didn't know that Dart doesn't have functions overloading 😢

Issue #384, opened 2020-07-15
Why Controller can't close


KeepFish:
This is page
```
import 'package:flutter/material.dart';
import 'truck_list_controller.dart';
import 'package:get/get.dart';

class TruckListPage extends StatelessWidget {

  final TruckListController c = Get.put(TruckListController());

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: Text('App bar'),
      ),
      body: Text('21321312'),
    );
  }
}

```
This is controller 
```
import 'package:get/get.dart';

class TruckListController extends GetxController {

  @override
  void onClose() {
    print('TruckListController onClose');
    super.onClose();
  }
}
```
When push in TruckListPage and pop. Print in console like 
![image](https://user-images.githubusercontent.com/12891782/87528918-d226cc80-c6c0-11ea-9c4d-fd7a1f15c494.png)
Why onClose method is not called.

KeepFish:
```
class TruckListPage extends StatelessWidget {

  @override
  Widget build(BuildContext context) {
    final c = Get.put(TruckListController());

    return Scaffold(
      appBar: AppBar(
        title: Text('App bar'),
      ),
      body: Text('21321312'),
    );
  }
}
```
When I put the Get.put in build, it's fine.....  Why?
![image](https://user-images.githubusercontent.com/12891782/87530080-75c4ac80-c6c2-11ea-8b89-19d5d6cc3dd8.png)

jonataslaw:
Please follow the steps for opening the issue, explaining data such as Get version and reproduction of the problem.
Both inside and outside the build the controller is being excluded in my test, so you must provide a case of reproducing the problem.

KeepFish:
Get version: 3.3.0
IDE: Android Studio 3.2.1 in macOS 10.14.5
Dart Version: 2.8.4

Reproduction:
Run 'flutter create myapp'
Then add some file like 
![image](https://user-images.githubusercontent.com/12891782/87618983-b4537900-c74d-11ea-95a8-460f84e9e0a2.png)
in main.dart
```
import 'package:flutter/material.dart';
import 'package:get/get.dart';
import 'second_page.dart';

void main() => runApp(GetMaterialApp(home: MyHomePage()));

class MyHomePage extends StatelessWidget {

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: Text('app'),
      ),
      floatingActionButton: FloatingActionButton(
[...]

PR #820, opened 2020-11-25
Adjustment in the sample code in the readme of all languages

eduardoflorence:
- **Change to the readme sample code, in the "Counter App with GetX" section**

**Before:**
```dart
class Home extends StatelessWidget {
  final Controller c = Get.put(Controller());
  @override
  Widget build(context) => Scaffold(
      ...  
```

**After:**
```dart
class Home extends StatelessWidget {
  @override
  Widget build(context) {
    final Controller c = Get.put(Controller());
    return Scaffold(
        ...
```

- **Motivation**: Avoid issues of beginners who do not use named routes. If Get.put is placed before the build method, onInit/onReady/onClose are not called and when navigating to another route the controller is not removed from memory

- **Referência**: https://github.com/jonataslaw/getx/issues/384#issuecomment-661919495

### DIFF
--- a/README-es.md
+++ b/README-es.md
@@ -125,11 +125,13 @@ class Controller extends GetxController {
 ```dart
 class Home extends StatelessWidget {
 
-  // Cree una instancia de su clase usando Get.put() para que esté disponible para todas las rutas "secundarias" allí.
-  final Controller c = Get.put(Controller());
-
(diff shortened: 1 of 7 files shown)


jonataslaw:
Thank you, you're right
LGTM

PR #836, opened 2020-11-30
Fix bugs in GetConnect

eduardoflorence:
- Request Method without decoder parameter (http.dart)
- Remove "final" of decoder parameter (request.dart)
- Type methods for standardization


jonataslaw:
Thanks for that.
LGTM!

jorelkimcruz:
Also when using get with dio. GetConnects Response<dynamic> is conflicting with dios own Response<dynamic>.

PR #851, opened 2020-12-03
Fix index out of range when parsing response

haidang93:
The parameter value variable has the same name with the original value causes error 'index out of range' when calling value[index] in the next lines.

### DIFF
--- a/lib/get_connect/http/src/response/response.dart
+++ b/lib/get_connect/http/src/response/response.dart
@@ -227,11 +227,11 @@ class HeaderValue {
           parameters[name] = null;
           return;
         }
-        var value = parseParameterValue();
-        if (name == 'charset' && value != null) {
-          value = value.toLowerCase();
+        var val = parseParameterValue();
+        if (name == 'charset' && val != null) {
+          val = val.toLowerCase();
         }
-        parameters[name] = value;
+        parameters[name] = val;
         bump();
         if (done()) return;
         if (value[index] == valueSeparator) return;
--- a/lib/get_connect/http/src/response/response.dart
+++ b/lib/get_connect/http/src/response/response.dart
@@ -227,9 +227,9 @@ class HeaderValue {
           parameters[name] = null;
           return;
         }
-        var val = parseParameterValue();
-        if (name == 'charset' && val != null) {
-          val = val.toLowerCase();
+        var valueParameter = parseParameterValue();
+        if (name == 'charset' && valueParameter != null) {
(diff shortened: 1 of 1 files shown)


eduardoflorence:
Wouldn't it be better for the variable name to be `valueParameter` instead of `val`?

haidang93:
> Wouldn't it be better for the variable name to be `valueParameter` instead of `val`?
anything unique will do : ) I just make a quick fix.

lib/get_connect/http/src/response/response.dart
```diff
@@ -227,11 +227,11 @@ class HeaderValue {
           parameters[name] = null;
           return;
         }
-        var value = parseParameterValue();
-        if (name == 'charset' && value != null) {
-          value = value.toLowerCase();
+        var val = parseParameterValue();
+        if (name == 'charset' && val != null) {
+          val = val.toLowerCase();
```
eduardoflorence, line 232:
```suggestion
        var valueParameter = parseParameterValue();
        if (name == 'charset' && valueParameter != null) {
          valueParameter = val.toLowerCase();
```

lib/get_connect/http/src/response/response.dart
```diff
@@ -227,11 +227,11 @@ class HeaderValue {
...
         }
-        parameters[name] = value;
+        parameters[name] = val;
```
eduardoflorence, line 234:
```suggestion
        parameters[name] = valueParameter;
```

eduardoflorence (changes requested):
Best variable name

eduardoflorence (approved):

lib/get_connect/http/src/response/response.dart
```diff
@@ -227,11 +227,11 @@ class HeaderValue {
           parameters[name] = null;
           return;
         }
-        var value = parseParameterValue();
-        if (name == 'charset' && value != null) {
-          value = value.toLowerCase();
+        var valueParameter = parseParameterValue();
+        if (name == 'charset' && valueParameter != null) {
+          valueParameter = val.toLowerCase();
```
eduardoflorence, line 232:
```suggestion
          valueParameter = valueParameter .toLowerCase();
```

eduardoflorence (changes requested):
sorry, a 'val' remained

eduardoflorence (approved):

haidang93:
perfect 🥇

jonataslaw:
Thanks for your contribution, LGTM

PR #859, opened 2020-12-05
GetResponsiveView

SchabanBo:
I think in this time we all need that widget, that's changes the view the screen size changes.
That's why i write this widget.
and its all here.
https://github.com/SchabanBo/getx#getresponsiveview
Thanks again.


jonataslaw:
Hi, thanks for you contribution!
I think that some PR code can be removed as it is already present in the framework.
responsiveValue is an example

https://github.com/jonataslaw/getx/blob/master/lib/get_utils/src/extensions/context_extensions.dart#L122

SchabanBo:
In the GetResponsiveView I can change the limits for the screen types through the settings property.
So if I used the method that is in the framework, the settings will not be considered and I will get a wrong result.
That way I had to rewrite it.

jonataslaw:
LGTM

PR #860, opened 2020-12-05
Demo test of invalid description of offUntil and offNamedUntil predicate

vbuberen:
This PR created to demonstrate that currently suggested description of predicate parameter in `offUntil` and `offNamedUntil` comments leads to wrong navigation stack. 

The quote from invalid comments:
```
/// [predicate] can be used like this:
/// `Get.until((route) => Get.currentRoute == '/home')`so when you get to home page,
```
Also, this PR appeared to support another PR which fixes mentioned issue.

Can be closed after checking.

PR #861, opened 2020-12-05
Update offUntil and offNamedUntil docs

vbuberen:
This PR changes docs for `offUntil()` and `offNamedUntil()` navigation extensions, because current description for the `predicate` parameter is invalid and leads to wrong navigation stack if used as suggested. I got caught by such invalid description on one project, so fixing to prevent others from incorrect usage.

To confirm that current description is wrong I updated tests to use the predicate in suggested way and opened #860 

After this PR gets merged I will create another PR to update navigation extensions tests to check exactly one behavior in one tests, because with this PR I test multiple conditions in one test.


jonataslaw:
Thanks!
LGTM!

PR #864, opened 2020-12-06
Handle for List field with native datatype

jasonlaw:
Need to handle for List with native datatype, such as List<String>.


jonataslaw:
LGTM

PR #871, opened 2020-12-07
Added will pop scope to defaultDialog

rakeshlanjewar:
Hello, I would like to add willpopscope to defaultdialog() , So that we can make loading dialog with ease.

Hope It will find community useful. 

Please add this kind of functionality where we can show loading dialog. Or let's discuss further regarding this

### DIFF
--- a/lib/get_navigation/src/extension_navigation.dart
+++ b/lib/get_navigation/src/extension_navigation.dart
@@ -314,6 +314,9 @@ extension ExtensionDialog on GetInterface {
     double radius = 20.0,
     //   ThemeData themeData,
     List<Widget> actions,
+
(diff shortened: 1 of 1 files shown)


rakeshlanjewar:
I know we can use dialog method but defaultdialog() is Love.

jonataslaw:
Thank you for your contribution, I find this very useful, but I need to check if it cannot cause side effects with the back() function.
After checking this I will accept this PR, thanks.

rakeshlanjewar:
Please check and merge as soon as possible, I have tested it, if back method called then only it will pop.

Thanks.

eduardoflorence (approved):

PR #872, opened 2020-12-07
Fix bug in GetHttpClient._createUri

reinaldowebdev:


jonataslaw:
Ow, I think I missed that.
Thank you for your contribution, lgtm

PR #890, opened 2020-12-10
solving problem of sending parameters by query parameters

rodrigorahman:
I found a problem in using GetConnect in the get method, when I send the query parameter it was not adding the url. I implemented the solution, feel free to analyze and see if it makes sense.

```dart
    final response = await get<List<GiphyModel>>(
      'http://api.giphy.com/v1/gifs/trending',
      query: {'api_key': 'chaveXXX'},
      decoder: (data) {
        return data['data']
            .map<GiphyModel>(
              (d) => GiphyModel(
                d['id'],
                d['images']['original']['url'],
              ),
            )
            .toList();
      },
    );
```

### DIFF
--- a/lib/get_connect/http/src/http.dart
+++ b/lib/get_connect/http/src/http.dart
@@ -77,11 +77,12 @@ class GetHttpClient {
     if (baseUrl != null) {
       url = baseUrl + url;
     }
-    final uri = Uri.parse(url);
-    if (query != null) {
-      return uri.replace(queryParameters: query);
+    if(query != null){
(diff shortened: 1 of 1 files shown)


eduardoflorence:
This bug has already been resolved with the commit below, but is not yet in pub.dev:

https://github.com/jonataslaw/getx/commit/1826a9f3f263af07dd3f0eb872e57adb5affcdbc#diff-21da8c9fcad48451c9aa7926f2df3277296857602228898fb5141536e87d2295

Duplicate of #872

PR #902, opened 2020-12-13
Getsocket allowselfsigned

eduardoflorence:
1. I made a Websocket Server in NodeJS to test `GetSocket`. I needed to set the URL to `ws://xxx.xxx` and for that I needed the `allowSelfSigned` property to be set to false, but it was fixed in true. Now I put it to be one of the parameters of `GetSocket`. 
P.S.: `GetSocket` worked very well.

2. When instantiating `GetSocket` the code editor only suggested the `close` method, since the others were not yet referenced in `BaseWebSocket (stub)`. I put everyone following the guidelines of the `close` method.


jonataslaw:
Hi, thanks for it
LGTM

PR #905, opened 2020-12-14
Translation of README for Indonesian

pratamatama:
Greetings from Indonesia! I'm Tamma, but I prefer being Haru so please call me Haru. 😄 
I'm a native speaker from Indonesia.

It's been a while for me since I start GetX-ing and fall in love with its simplicity and how it separate things from the view.

In the last 3 days I have spare time to spend and tried to refresh my mind by re-reading the docs, while doing that, I was thinking about how wonderful it is to have a localized version of the docs so I decided to do both read and contribute by translating it to our language.

In this PR, I added Indonesian translation for all available documentation started and translated from the main README file.

This is also my first contribution to the open source project, so please let me know if something needs to be adjusted. Thank you very much for making such an incredible package that makes our work much easier!


jonataslaw:
Hi, thank you very much for your contribution, Indonesia has a fantastic community of Flutter, and having documentation in the native language is certainly something much more comfortable.
LGTM

Issue #916, opened 2020-12-16
test failed


ValeriusGC:
trying to test in example raised error:

![Screenshot from 2020-12-16 13-48-03](https://user-images.githubusercontent.com/3320861/102325861-7d57ec80-3fa5-11eb-8f09-13a2d73e5c57.png)

SchabanBo:
It works now :)

jonataslaw:
Solved by https://github.com/jonataslaw/getx/pull/917

PR #917, opened 2020-12-16
Fix example test #916

SchabanBo:
When we updated the example to nested route, we forgot to update the test too.
#916


eduardoflorence (approved):

jonataslaw:
Thanks! 
I forgot to update the example with the child pages
LGTM

PR #920, opened 2020-12-16
Body string for content-type application/x-www-form-urlencoded

eduardoflorence:
To make a request with content-type `application/x-www-form-urlencoded` it is necessary that the body be a string and GetConnect did not accept the body as a string.
Example:
```dart
post(
  'https://api.example.com.br/auth',
  'user=eduardo&password='123',
  contentType: 'application/x-www-form-urlencoded',
);
```


jonataslaw:
Thanks! 
LGTM

PR #925, opened 2020-12-18
Add "empty" status to rx_notifier as in many cases we need to show empty animation/image/...  instead of a list

alizera:
In many cases we need to show a particular widget when the response object has an empty list, In the real world `responseObject` is a kind of object that describes the fetched data's schema and one of received data might be a list that can be null or empty, and in such situation, we can show an animation, image or widget to inform user there is no contact/order/etc... 
Like:

4-empty-state-mobile-app-designs

I've added **isEmpty** status to handle this situation more easily (instead of checking data inside widget, we check it inside controller).


jonataslaw:
Hi, thanks for your contribution. 
Looks good to me. 
I'll merge

PR #928, opened 2020-12-18
Add plural in get Internationalization.

CpdnCristiano:


jonataslaw:
Nice!
Can you provide a quick example just to add to the documentation?
It can be just one line, just demonstrating how to use it.

CpdnCristiano:
I just added the example

jonataslaw:
Thanks! LGTM

RafaRuiz:
I believe this cannot be compared to the plurals that, for example Android has.
The implementation I see is if `i > 1`, but in most of the languages, I say:
- 0 Results
- 1 Result
- 2 Results.
This instead will return:
- 0 Result
- 1 Result
- 2 Results.
That's not the only issue that I see, the plurals can be more than "more than one". In several languages, there are N ways to write something, where N can depend if the quantity is: "zero", "one", "few", "many", or "two) (reference: Android)
image

I believe the `keys` in `Translations` should handle only the "non quantity-related" localisations and there should be another abstract Map that handles all these possibilities.

If this is an issue, I can send work on it and send a PR.

jonataslaw:
> I believe this cannot be compared to the plurals that, for example Android has.
> The implementation I see is if `i > 1`, but in most of the languages, I say:
> 
> * 0 Results
> * 1 Result
> * 2 Results.
>   This instead will return:
> * 0 Result
> * 1 Result
> * 2 Results.
>   That's not the only issue that I see, the plurals can be more than "more than one". In several languages, there are N ways to write something, where N can depend if the quantity is: "zero", "one", "few", "many", or "two) (reference: Android)
>   image
> 
> I believe the `keys` in `Translations` should handle only the "non quantity-related" localisations and there should be another abstract Map that handles all these possibilities.
> 
> If this is an issue, I can send work on it and send a PR.

Nice note! 
If you can contribute with it, please, send a PR

RafaRuiz:
it would involve a breaking change in the plurals though, it wouldn't accept a "plural key", and I'd have to create an abstract `pluralKeys` (in adition to the `keys`). What do you think? I'll start it now and will discuss it in the meantime!

RafaRuiz:
Ok, I found out something. 
The currrent function for plurals fails in all the languages, no exception.
To make this right at 100%, we would need to follow the unicode plurals rules:
https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html

To do that, we would have to include the "intl" library that has a full implementation of it. But I'm not sure if that's possible in this project. Can you please confirm? in that case, we could also fix the formatting (at the moment it doesn't support `%d` formatting either).

If so, I'll start working on it, otherwise, it would be so bad that we couldn't use it, the plural stuff has no use at all.

jonataslaw:
I prefer to keep the library free of additional dependencies. I think we will have a long work ahead of us to implement this.

RafaRuiz:
Perfect, that sounds great @jonataslaw .
What I can do though is a small introduction, and making the PLURALS valid for the most spoken languages.
I believe with `one` and `other`, ENGLISH, SPANISH and CHINESE are well covered.
The current implementation doesn't match either of those.
What do you think? At least we could cover most of the speakers and leave a WIP for the rest of them.
Let me know and I'll start moving my hands! tap tap tap :)

jonataslaw:
In fact I believe that the algorithm was made for Portuguese, which has the same rules as Spanish, so I believe that there are no break changes if we change the algorithm to "== 1" instead of "> 1".
I think it was a typo, it can be fixed.

RafaRuiz:
sure, here you go then
#1085

PR #943, opened 2020-12-23
Another translation option with parameters

CpdnCristiano:
- trArgs is not very intuitive, and not very useful when you have to pass several parameters.
- trParams use the map, it is more intuitive to pass parameters
```dart


Map<String, Map<String, String>> get keys => {
    'en_US': {
        'logged_in': 'logged in as @name with email @email',
    },
    'es_ES': {
       'logged_in': 'iniciado sesión como @name con e-mail @email',
    }
};

Text('logged_in'.trParams({
  'name': 'Jhon',
  'email': 'jhon@example.com'
  })); /// logged in as Jhon with email jhon@example.com
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,27 @@ var products = [];
 Text('singularKey'.trPlural('pluralKey', products.length, Args));
 ```
 
+#### Using translation with parameters
+
+```dart
+import 'package:get/get.dart';
+
+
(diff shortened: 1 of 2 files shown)


jonataslaw:
This is really incredible, I'm afraid this will cause some conflict in the future, we need to think of a way to escape @variable in the future. For now, LGTM
Thanks for this!

PR #954, opened 2020-12-28
GetMiddleware - Redirect: get arguments

eduardoflorence:
The redirect method of GetMiddleware, received only the name of the route, so I modified it to also receive the `arguments` passed in navigation for the route, so we can pass these arguments also to the redirected route or make conditions with it. Example:

```dart
class CheckArgumentsMiddleware extends GetMiddleware {
  @override
  RouteSettings redirect(String route, {Object arguments}) {
    return arguments == 'test'
        ? RouteSettings(name: '/login/test', arguments: arguments)
        : RouteSettings(name: '/login/prod', arguments: arguments);
  }
}
...
GetPage(
  name: '/login',
  page: () => LoginPage(),
  binding: LoginBinding(),
  middlewares: [CheckArgumentsMiddleware()],
),
...
Get.toNamed('/login', arguments: 'test');
...

```

### DIFF
--- a/lib/get_navigation/src/routes/route_middleware.dart
+++ b/lib/get_navigation/src/routes/route_middleware.dart
@@ -30,7 +30,7 @@ abstract class _RouteMiddleware {
   /// }
   /// ```
   /// {@end-tool}
-  RouteSettings redirect(String route);
+  RouteSettings redirect(String route, {Object arguments});
 
   /// This function will be called when this Page is called
   /// you can use it to change something about the page or give it new page
@@ -81,7 +81,7 @@ class GetMiddleware implements _RouteMiddleware {
   GetMiddleware({this.priority});
 
   @override
-  RouteSettings redirect(String route) => null;
+  RouteSettings redirect(String route, {Object arguments}) => null;
 
   @override
   GetPage onPageCalled(GetPage page) => page;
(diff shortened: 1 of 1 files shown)


jonataslaw:
This is apparently a break update, so unfortunately it cannot be part of a version 3 update.
Just for the sake of conscience: Get.arguments is not available within Middleware? Because if it is already filled, there would be no need to pass it as an argument

eduardoflorence:
> This is apparently a break update, so unfortunately it cannot be part of a version 3 update.
> Just for the sake of conscience: Get.arguments is not available within Middleware? Because if it is already filled, there would be no need to pass it as an argument

I did a test now and `Get.arguments` returns null in GetMiddleware.

Even putting the arguments parameter as optional, the signature of the redirect method that needs to be overwritten must have the parameter declaration and that is why the breaking change happened. Since this feature is still very recent, perhaps the change would not impact many users if we tell you in the changelog, but I understand your concern if it is not possible.
We recently had a question in a developer group that would like to always redirect to an error page if the arguments were empty and I also saw that we were unable to pass the arguments to the new page that will be redirected.

shmy:
any update?

miaosun009:
@jasonlaw Need this

miaosun009:
Will this feature never appear? @eduardoflorence

appinteractive:
@jonataslaw Any plans to add this basic feature? Would be very appreciated.

zeewell:
Need this FEATURE very much! @jonataslaw

appinteractive:
I solved it by routing with a service that decides what to do instead of a middleware.

PR #962, opened 2020-12-29
GetMiddleware: Set Get.Parameter after runOnPageCalled

eduardoflorence:
Get.parameter must be set after calling runOnPageCalled, as page can be modified. Example:

```dart
@override
GetPage onPageCalled(GetPage page) {
  return authController.username != null
      ? page.copyWith(parameter: {'user': authController.username})
      : page;
}
```

```dart
final runner = MiddlewareRunner(match.route.middlewares);
route = runner.runOnPageCalled(match.route);
Get.parameters = route?.parameter;  // This PR
```

### DIFF
--- a/lib/get_navigation/src/routes/route_middleware.dart
+++ b/lib/get_navigation/src/routes/route_middleware.dart
@@ -206,6 +206,7 @@ class PageRedirect {
   /// check if redirect is needed
   bool needRecheck() {
     final match = Get.routeTree.matchRoute(settings.name);
+    Get.parameters = match?.parameters;
 
     // No Match found
     if (match?.route == null) {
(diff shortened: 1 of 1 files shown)


jonataslaw:
Nice! Thanks for this PR

eduardoflorence:
After not passing the test, I verified that `Get.Parameters` only gets its value from dynamic urls. I changed this PR to allow the value of the `parameter` property of `GetPage` to be added to `Get.Parameters`

jonataslaw:
LGTM

BrianMwas:
Hey, I want to add auth check on a start screen to prevent the app from going to the login. Where do I add these middlewares?

PR #969, opened 2020-12-31
FIX: Post headers in GraphQL (query & mutation)

asalvi0:
In the GraphQL methods implementation the headers are not in use, this commit just adds the headers param to the post.


jonataslaw:
Thanks for that, LGTM!

PR #998, opened 2021-01-10
Added lang directory that contains the translations_service class and two languages, english and portuguese, configured to use the device system language and use the english like default case the device language is'nt listed on the lang files.

rodriguesJeff:


jonataslaw:
Indeed, the standard example had no sample of internationalization. Thank you for that. LGTM

PR #1005, opened 2021-01-11
Snackbar run Get.back multiple times if you tap to dismiss multiple times

### LINKED ISSUE #997: pop current route when double tap on screen while snackbar is visible
void getSnackBar({String title, String msg}) {
  Get.snackbar(
    title,
    msg,
    margin: EdgeInsets.all(10),
    forwardAnimationCurve: Curves.fastLinearToSlowEaseIn,
    reverseAnimationCurve: Curves.fastLinearToSlowEaseIn,
    duration: Duration(milliseconds: 1000),
    backgroundGradient: myLinearGradient,
    // isDismissible: false,
    // borderRadius: 4,
    overlayBlur: 2,
    snackPosition: SnackPosition.BOTTOM,
    instantInit: true,
  );
}

pop current page when double tap on the screen while snackbar is visible.
it would be better if just snackbar get off, doesn't matter how many time we have tapped on screen.

eduardoflorence:
Prevents Get.back from being executed more than once if the user clicks the screen several times to dismiss the snackBar

Fixes #997

### DIFF
--- a/lib/get_navigation/src/snackbar/snack_route.dart
+++ b/lib/get_navigation/src/snackbar/snack_route.dart
@@ -50,6 +50,7 @@ class SnackRoute<T> extends OverlayRoute<T> {
   Alignment _initialAlignment;
   Alignment _endAlignment;
   bool _wasDismissedBySwipe = false;
+  bool _onTappedDismiss = false;
 
   Timer _timer;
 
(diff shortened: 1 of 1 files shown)


jonataslaw:
Thanks for this contribution!
LGTM!

PR #1067, opened 2021-02-11
params in url by Get.parameters

enghitalo:
this permit this kind of syntax
`Get.parameters.addAll({"name": "varginia", "vinis": "viniiss"});
Get.toNamed("/second", arguments: 'Get is the best');

print(Get.parameters); // {name: varginia, vinis: viniiss}`
//result in urls => http://localhost:32773/#/second?name=varginia&vinis=viniiss

### DIFF
--- a/lib/get_navigation/src/extension_navigation.dart
+++ b/lib/get_navigation/src/extension_navigation.dart
@@ -541,25 +541,24 @@ extension GetNavigation on GetInterface {
     String page, {
     dynamic arguments,
     int id,
+    Map<String, String> parameters,
     bool preventDuplicates = true,
   }) {
-    Get.parameters.forEach((key, value) async {
-      if (await page.contains('?')) {
-        page = "$page${"&$key=$value"}";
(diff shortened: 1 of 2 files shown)


jonataslaw:
I was going to implement exactly that function now!
Thank you for that.
I just suggest that this function be a navigation parameter.

enghitalo:
@jonataslaw happy to ear it. The project is 100% done I needed create a new test and change the syntax;
I thing that syntax is more consintent now, tho. XD
      ```
 var parameters = <String, String>{
        "varginias": "varginia",
        "vinis": "viniiss"
      };
      parameters.addAll({"a": "b", "c": "d"});
      Get.toNamed("/buda", parameters: parameters);

     print(Get.parameters); // {name: varginia, vinis: viniiss}
```

jonataslaw:
Very nice! 
It looks like you got it out of my mind.
I'll merge, don't worry, if there is a problem I'll adjust it.

PR #1074, opened 2021-02-13
fixed the log when using builder instead of widget on Get.to().

IsmailAlamKhan:
When routing using the builder instead of the widget on Get.to the Get.log was printing the builder instead of the name of the widget.


jonataslaw:
This really solves the problem, however, it looks like the widget builder will be called twice in this situation

IsmailAlamKhan:
@jonataslaw any workaround for that? And B.T.W thank you soo much for this awesome package.

PR #1085, opened 2021-02-18
Make plurals happen with 0 as well.

RafaRuiz:
Example:
- 0 Results.
- 1 Result.
- 2 Results.


jonataslaw:
LGTM, Thanks!

WolfVic:
I'm not sure if it's the right place, but this change is not working... It returns plurals only when there is 1 item.  We need to flip the expressions or modify the condition to `i != 1`

PR #1092, opened 2021-02-20
Fixed bug when bodyBytes is null

jasonlaw:
When body is empty, the bodyBytes will be null, hence causing following null error:
E/flutter (28391): [ERROR:flutter/lib/ui/ui_dart_state.cc(177)] Unhandled Exception: NoSuchMethodError: The getter 'length' was called on null.

Fixed by null checking.


jonataslaw:
Thanks, LGTM

PR #1134, opened 2021-03-01
Create an Event Rx holder to call listeners when the same value is passed

RafaRuiz:
This idea comes half-way from the [SingleLiveEvent of Android and Kotlin](https://medium.com/@abhiappmobiledeveloper/android-singleliveevent-of-livedata-for-ui-event-35d0c58512da), although this one doesn't keep the value afterwards (maybe if it's interesting, I can make it work in both ways!); this new `RxEvent<T>` will notify the listeners through `RxEvent<T>.trigger(T)`.

Imagine you're creating an `Obx(builder)` where the builder animates a Widget for `2` seconds, and you want to animate that according to a logic variable: `RxEvent<int> secondsAnimation = RxEvent<int>(2)`. After an user iteration, you might want to animate that again to 2 seconds as well: `secondsAnimation.trigger(2)` will trigger the listeners again (`.call(2)` or `.value = 2` won't!). 

I'm up to hear some feedback, or event remove entirely the `RxEvent<T>` class and add the `.trigger(T)` method into `Rx<T>`.

### DIFF
--- a/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
+++ b/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
@@ -379,6 +379,30 @@ class Rx<T> extends _RxImpl<T> {
   }
 }
 
+/// Similar class to Rx<T> but this also will refresh the listeners if the same
+/// value has been provided. This is useful when maintaining a state with the
+/// same user = User("foo").
+/// For example, supposed we have a `int seconds = 2` and we want to animate
+/// from invisible to visible a widget in two seconds:
+/// RxEvent<int>.call(seconds);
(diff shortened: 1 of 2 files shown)


jonataslaw:
Interesting feature, but I was thinking, trigger could be part of the core instead of having its own Rx for that. It could be in _RxImpl to allow it to be used in existing Rx variables

RafaRuiz:
Thanks for the answer. I'll update that!

RafaRuiz:
All changed now @jonataslaw

jonataslaw:
Thanks for your contribution, LGTM!

PR #1142, opened 2021-03-03
Revert "Fixed bug when bodyBytes is null"

jonataslaw:
Reverts jonataslaw/getx#1092

This is causing problems with GetConnect, reverting to fix


jasonlaw:
Hi @jonataslaw ,
What is the right fix then? 
The bodyBytes.length will still causing the null error when body is empty.

Thanks.

PR #1164, opened 2021-03-08
Get.put/putAsync/find not nullable

### LINKED ISSUE #1156: Get.put<T> has it possible to return null ?
`get: ^4.0.0-nullsafety.2`

`final UserController controller = Get.put(UserController())!`

when use `Get.put`, must add "!", Is it possible to omit?

and 

`final Rx<String> title = Rx<String>('Hello Word!');`

I've given the default value, If I want to use title value, must to add "!" or "?", Is it possible to omit?

eduardoflorence:
- I did a sequence of tests and it was possible to make `Get.put`, `Get.putAsync` and `Get.find` as **non-nullable**.
- I modified the test files to treat the above methods as non-nullable

Fixes #1156


jonataslaw:
Thanks, that was what was missing to launch null-safety as stable

Mayb3Nots:
hi @jonataslaw is this on stable already? It weird to make sure` Get.put() `is not null every time.

PR #1223, opened 2021-03-20
Little improvements of example code

3lB4rt0:
It's my first pull request, i'm not sure i did it well.

- I removed BoxDecoration because I think it doesn't affect on anything.
- I replaced GetX<ControllerX> with Obx, because like there is wrote [here](https://github.com/jonataslaw/getx/blob/master/documentation/en_US/state_management.md#getbuilder-vs-getx-vs-obx-vs-mixinbuilder) (if I understand correctly) it consume less RAM and use the controller already instantiated with GetView

PS: I really love Getx. It's so simple!


jonataslaw:
Thanks for your contribution, 
LGTM

PR #1250, opened 2021-03-29
Implement own .call for Rxn variables

RafaRuiz:
`Rxn` variables accept `null` as a possible state but when calling `.call(null)` the value doesn't either update or trigger the callbacks.


lib/get_rx/src/rx_types/rx_core/rx_string.dart
```diff
@@ -286,4 +286,13 @@ class RxnString extends Rx<String?> implements Comparable<String>, Pattern {
   int compareTo(String other) {
     return value!.compareTo(other);
   }
+
+  String? call([String? v]) {
+    if (v == null) {
+      value = null;
```
jonataslaw, line 292: Okay, here we have something to think about.
call is used as a getter and as a setter, and it cannot set a value to null when the String has no value, otherwise the user will try to get a getter and it will clean up its Rx instead. And this is something that we will have to investigate how to manage, because in theory the user can also set null to change the value. It may not allow a null value, and return to the "nil ()" method to assign null to Rx. IDK

RafaRuiz: Hey there, thanks for the comments.
I think all I can say, if you take it from the abstract theory of why the `Rxn` started existing, it's because we can now create a `Rx` that hasn't "started" or "isn't ready" yet, meaning that we can use a loading state to true, false or "I'm not even initialized yet".
But there's also a concept of **resetting** a state, and we might want to reset a certain variable.
Think of it also as a TextField validation. You want to validate some text to VALID or NOT_VALID, but when there's no text, you might not want to valid as "there's nothing yet to validate". This is also a common practice, you're not telling your users "Invalid email" when the field is invalid.

It was my main thought :)

RafaRuiz: I'm not under another situation.
I have a `ListView` in a screen, and I have a `final RxnString idVideoPlaying`, which holds the video that is playing at the moment, something like a facebook videos feed. I want to stop the video playing, so making it `null` would be ideal.

PR #1257, opened 2021-03-29
better error message for navigation

maxzod:
it took me almost three hours trying to figure out why?
in the end the reason was that 
`Get.to() ` and  `Get.toNamed()`
both accept Strings
and the error message is already blurry enough !


jonataslaw:
Thanks, it looks better
LGTM

PR #1262, opened 2021-03-30
Fix for darkTheme not getting changed correctly if default value is set in GetMaterialApp

GoldenSoju:
Simple fix to avoid that ThemeData doesn't get changed correctly when standard value for darkTheme is set in GetMaterialApp and Theme should be changed to a ThemeData with Brightness.dark.

To test this issue, you would need two ThemeData with Brightness.dark and try to switch between those two with Get.changeTheme  **AND** have a default value for dark Theme in GetMaterialApp.

```
runApp(
    GetMaterialApp(
      themeMode: ThemeMode.system,
      theme: CustomThemes.lightStandardTheme,
      darkTheme: CustomThemes.darkStandardTheme,
    ),
  );
  ```

ThemeChanges work perfectly if you have **NO** default value set for darkTheme, but I don't think this is the correct behaviour.


jonataslaw:
Thanks for this PR. 
LGTM

Issue #1267, opened 2021-04-01
Unable to use `Get.reload()` and `Get.reloadAll()`


lkloon123:
**Describe the bug**
`Get.reload()` and `Get.reloadAll()` isn't defined for type `_GetImpl`

**Reproduction code
NOTE: THIS IS MANDATORY, IF YOUR ISSUE DOES NOT CONTAIN IT, IT WILL BE CLOSED PRELIMINARY)**

example:

```dart
void main() {
   Get.reload();
   Get.reloadAll();
}
```

**To Reproduce**
just go to anywhere and type in `Get.reload()` and the error shown

**Expected behavior**
Should able to use without error

**Screenshots**
<img width="647" alt="Screenshot 2021-04-01 at 22 19 02" src="https://user-images.githubusercontent.com/21114981/113308582-e5ee8d80-9338-11eb-9119-cbf84160aeb1.png">

**Flutter Version:**
`2.0.1`

**Getx Version:**
`3.26.0`

PR #1268, opened 2021-04-01
Add missing reload() and reloadAll()

lkloon123:
Simple fix for #1267


jonataslaw:
Ow, thanks for that. 
LGTM

4F2E4A2E:
What's the actual use case here?

PR #1273, opened 2021-04-01
[GetConnect] defaultDecoder can be null

Goddchen:
Only in `_get(...)` the `defaultDecoder` is cast to `Decoder<T>`. Everywhere else, e.g. in `_request(...)`, it is correctly cast to `Decoder<T>?`.


jonataslaw:
Thanks! LGTM

Issue #1275, opened 2021-04-03
../../../.pub-cache/hosted/pub.flutter-io.cn/get-4.1.2/lib/get_navigation/src/extension_navigation.dart:1107:40: Error: A value of type 'Locale?' can't be returned from a function with return type 'Locale' because 'Locale?' is nullable and 'Locale' isn't.  - 'Locale' is from 'dart:ui'.   Locale get deviceLocale => ui.window.locale;


YugeCse:
版本：get: ^4.1.2
编译时报错如下：
../../../.pub-cache/hosted/pub.flutter-io.cn/get-4.1.2/lib/get_navigation/src/extension_navigation.dart:1107:40: Error: A value of type 'Locale?' can't be returned from a function with return type 'Locale' because 'Locale?' is nullable and 'Locale' isn't.
 - 'Locale' is from 'dart:ui'.
  Locale get deviceLocale => ui.window.locale;

  Locale get deviceLocale => ui.window.locale; 
  应该修改为 
  Locale? get deviceLocale => ui.window.locale;

nickwri:
I have the same issue. You can update the code in the package to fix this, but this is not a solution:

  Locale? get deviceLocale => ui.window.locale;

My project is very basic and runs fine in Android.

zpisgod:
same here, create a brand new project and run with 
`flutter run -d chrome`

YugeCse:
resolved

PR #1277, opened 2021-04-03
Changed deviceLocale getter to return Locale?

nickwri:
Not sure if this is the right solution to issue #1275, but added it for you to assess.

Stack issue raised here: https://stackoverflow.com/questions/66931450/devicelocale-error-in-simple-getx-flutter-project-when-running-on-web


jonataslaw:
Thanks, LGTM

PR #1288, opened 2021-04-06
Construct request with body for form-urlencoded.

aramayyes:
I am unable to make a post request with `contentType: 'application/x-www-form-urlencoded'` using GetConnect.
```dart
class AuthenticationProvider extends BaseProvider { // This just adds some headers.
  static const String _tokenEndpoint = 'Token';

  @override
  void onInit() {
    super.onInit();

    httpClient.defaultDecoder = (val) => TokenDto.fromJson(val);
  }

  // This request fails, because sent body is invalid for given content type.
  Future<Response<TokenDto>> getToken(SignInWithPasswordDto payload) => post(
        _tokenEndpoint,
        payload.toJson(),
        contentType: 'application/x-www-form-urlencoded',
        headers: {'Authorization': getConnectionConfig().authorization},
      );
}
```

So, the problem is in body constructing. Body should be constructed as a url string and not json.

### DIFF
--- a/lib/get_connect/http/src/http.dart
+++ b/lib/get_connect/http/src/http.dart
@@ -103,18 +103,20 @@ class GetHttpClient {
       headers['content-length'] = bodyBytes.length.toString();
       headers['content-type'] =
           'multipart/form-data; boundary=${body.boundary}';
+    } else if (contentType != null && contentType.toLowerCase() == 'application/x-www-form-urlencoded' && body is Map) {
+      var parts = [];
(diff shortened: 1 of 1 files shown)


jonataslaw:
Wow, thanks for your contribution, I inserted a TODO to check this out, but good thing you already did.
LGTM

Galti:
@aramayyes nice to see that you have solved this issue. Thanks!
@jonataslaw please, merge this PR ASAP. Thanks in advance.

jonataslaw:
It Is on pub now

PR #1327, opened 2021-04-17
feat:  extensions methods convert value in percent value

kauemurakami:
extensions methods convert value in percent value

### DIFF
--- a/lib/src/responsive/size_percent_extension.dart
+++ b/lib/src/responsive/size_percent_extension.dart
@@ -1,0 +1,9 @@
+import '../get_main.dart';
+
+//Converts a double value to a percentage
+extension PercentSized on double {
+  // height: 50.0.hp = 50%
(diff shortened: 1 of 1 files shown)


Ascenio:
Wouldn't it be better to make the extension receive a value of how much % do you want? Like this:
```dart
double get hp(double percent) => percent * (Get.height * (this / 100));
```
Because most of the time we are not interested in the value of `1%`, but rather some `value%`

Ascenio:
Also could you please add tests? I've found [this](https://stackoverflow.com/a/53707065) resource about how to mock the screen size in a Widget test.

kauemurakami:
just add the file that I inserted separately for this resource, percent_sixe_extension, guess in your project and use that way:
Container(height: 50.0.hp, width: 50.0.wp, color: Colors.red)

Ascenio:
What?

Issue #396, opened 2020-07-17
Question: How to pop all open dialogs?


Prn-Ice:
**Describe the bug**
Sorry if this question has an obvious answer but I want to know if there's any current way to pop all open dialogs.

jonataslaw:
Get.until((_)=> Get.isDialogOpen );

Prn-Ice:
Thanks.

PR #1369, opened 2021-04-28
Compatibility with dart-lang/mockito

lifez:
Related to [GenerateMocks not working on class that GetxController #396
](https://github.com/dart-lang/mockito/issues/396)

getx need to change some codebase to be compatible with [Mockito](https://github.com/dart-lang/mockito)


jonataslaw:
Thanks for this, LGTM

Issue #1422, opened 2021-05-11
GetConnect chinese garbled


Victor-XHH:
I try to use get4.1.4 GetConnect, the Chinese characters obtained from the api are garbled, how to set it in the defaultDecoder? Is it useful?

roipeker:
You might try to decode the `Response.bodyBytes` with utf8
The defaultDecoder is just a callback to transform the expected output in `Response.body` to a specific Type.

jonataslaw:
fixed on https://github.com/jonataslaw/getx/pull/1424

PR #1424, opened 2021-05-12
changed response default encoding from latin1 to utf8

heftekharm:
Default latin1 encoding causes unreadable characters when the language is not English . In my case, it was a Persian application  (The Persian language characters are almost the same in the Arabic language). For some days Our team was searching for the reason why this weird thing is happening and texts don't show properly in the application.

### DIFF
--- a/lib/get_connect/http/src/response/response.dart
+++ b/lib/get_connect/http/src/response/response.dart
@@ -60,14 +60,13 @@ class Response<T> {
   final T? body;
 }
 
-Future<String> bodyBytesToString(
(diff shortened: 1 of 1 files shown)


jonataslaw:
Thank you for that. I believe that this issue is also related to this: https://github.com/jonataslaw/getx/issues/1422

Perhaps this affects Latin languages, such as Spanish, Portuguese and French.

I'll do a quick test to make sure

anchao-lu:
How to set charset?

MarlonJD:
This changes is not on the pub.dev packages. I checked please fix this issue. on master response.dart and pub.dev source is different

PR #1451, opened 2021-05-20
Make withCredentials configurable

jasonlaw:
In order for cookie to work for flutter web, we need to set the withCredentials = true.


jonataslaw:
Thanks for this PR.
LGTM

PR #1452, opened 2021-05-20
Add content type and length into the headers when the content type is 'application/x-www-form-urlencoded'

calvingit:
Add 'content-type' & 'content-length' into the headers when the content type is 'application/x-www-form-urlencoded'


jonataslaw:
I think these headers are added automatically when using formdata.
But it may be that the dev just wants to send the map, without wrapping it in a FormData, and I think that solves this problem.
I need to be sure that setting this up cannot harm multipart file uploads, and then I can merge.
Thank you for that

PR #1457, opened 2021-05-22
Bug fixed: Redundant concatenating base url

jasonlaw:
Concatenating url with base url will be done inside _requestWithBody method. Removing from here as it is redundant.


jonataslaw:
Hi Jason, thanks for that. But I was wondering, what if the user wants to use the baseUrl for graphql?

jonataslaw:
Sorry, reading your issue I understand, baseUrl is merged into the url later, so this prevents it from being done twice

jonataslaw:
Thanks for this contribution, LGTM

PR #1459, opened 2021-05-23
GraphQLResponse inherit Response info

jasonlaw:
After a successful graphql call, the GraphQLResponse, which is inherited from Response, does not contains all the info, except body, from the query/mutation response. This makes all the current attributes such as hasError, isOk are not usable.

Fixed to fill up the response information when transforming to GraphQLResponse.


jonataslaw:
LGTM, thanks!

PR #1475, opened 2021-05-26
Fix anchor to heading not work on Vietnamese and some typo

inuyashaaa:
Fix anchor to heading not work on Vietnamese and some typo


jonataslaw:
Thanks! LGTM

PR #1496, opened 2021-05-29
Create getConnect.md

babakoto:
### GetConnect
GetConnect is an easy way to communicate from your back to your front with http or websockets (Not support socket io )
### Default configuration
You can simply extend **GetConnect** and use the GET/POST/PUT/DELETE/SOCKET methods to communicate with your Rest API or websockets.

```
class StoreProvider extends GetConnect {

}
```
### Configuration
```
class StoreProvider extends GetConnect {
 @override
  void onInit() {
    super.onInit();
    httpClient.baseUrl = "http://192.168.1.12:3000";
  }
}
```

### defaultDecoder
All request will pass to jsonEncode . Don't use default decoder if you have more than one type of model in your provider.

**Provider**
```
class StoreProvider extends GetConnect {
 @override
  void onInit() {
    super.onInit();
    httpClient.defaultDecoder = (data) => Store.formJson(data);
    httpClient.baseUrl = "http://192.168.1.12:3000";
  }
}
```
**Model**
```
class Store {
   String id;
   String name;

   Store.fromJson(Map json):
      this.id = json["id"],
      this.name = json["name"]
}
```

### DIFF
--- a/documentation/en_US/getConnect.md
+++ b/documentation/en_US/getConnect.md
@@ -1,0 +1,1 @@
+GetConnect


babakoto:
init documentation for GetConnect

PR #1497, opened 2021-05-30
[WIP] Implement Router outlet pattern for getx

ahmednfwela:
This PR simplifies nested navigation by making it a single navigator with a single navigation stack, and instead rendering the ui differently based on the current route.

I am still studying the best approach for this, but the general concept is to add a RouterOutlet widget as a place holder for the nested navigator, that would then listen to the router delegate and react to it to display different ui elements.

Some challenges are:

- [x] Decoupling the logical routes from the visual routes 
  - The most important challenge, basically how to navigate to a new page without the navigator taking over
  - How this was done is very straight forward, we filter what pages we pass to the `Navigator` inside the router delegate's build method
  ```dart
  List<GetPage> getVisiblePages() {
    return routes.where((r) {
      final mware =
          (r.middlewares ?? []).whereType<RouterOutletContainerMiddleWare>();
      if (mware.length == 0) return true;
      return r.name == mware.first.stayAt;
    }).toList();
  }  
  ```
- [x] Make Get's routing system work without the navigator (make it based on logical routes instead of visual routes)
  - Currently, Get relies too much on the `Navigator` and the events defined in `GetPageRoute`, but in order to make this new approach complete, we need to change how the bindings and middleware work.
  - This was done by mapping each `GetPage` to its related `GetPageRoute` in the router delegate and using the methods defined in `GetPageRoute` to emulate the old behavior.

- [ ] adapt the old navigation api defined in getx (e.g. `Get.toNamed`, `Get.to`, `Get.back`,...) to the new routing api.


jonataslaw:
Amazing job.
Let me know if you need something

jasonlaw:
May I know if this based on Navigator 2.0?

ahmednfwela:
@jonataslaw thanks for your great work on this library and simplifying our work flow 😄 

1. yes, this is based on the navigator 2.0 api to make it compatible with the web and the other new features
2. currently this is what the `example_nav2` can do:

ahmednfwela:
currently there are some conflicts between the way getx handles navigation based on the old api and the new router delegate api.
e.g. the way bindings/parameters/arguments work

Azzeccagarbugli:
This is going to be really useful. Thanks a lot, looking forward for the merging 🚀

ahmednfwela:
I managed to improve how the routing works  and now it's -almost- identical to how Getx works now.

I still don't know how to emulate the transitions between routes in a `RouterOutlet`, since navigator used to handle that automatically (and I am not an expert with animations), so if someone can help that would be great



Example src code: https://github.com/Bdaya-Dev/getx/tree/router-outlet/example_nav2

Azzeccagarbugli:
I think that should be necessary to update also the whole section in the README about the routing @ahmednfwela, do you need any help?

ahmednfwela:
That would be great, but I am still not sure how to adapt the old api

- Old API has a single navigation stack and you append to it if you want to move to a new route
  e.g. going from route A to B :
  `/A => /A/B`
- New API has 2 navigation stacks (browser history + route identifier)
  e.g.  going from route A to B :
  `/A => /B`

Do we make Getx use the new api ALWAYS ? or only if the user opts in to the new Navigator API (by setting the router delegate) ?

Azzeccagarbugli:
> That would be great, but I am still not sure how to adapt the old api

Me neither unfortunately

> Do we make Getx use the new api ALWAYS ? or only if the user opts in to the new Navigator API (by setting the router delegate) ?

I think that Getx should use the new API always, and for this reason I think that this PR could be a major release. Flutter is now officially stable on the web too so the new standard should provide a nice way to handle routing, and this navigator update can fix the gap.

----

Anyway I think that we should wait an answer from @jonataslaw and see how to move but meanwhile amazing job @ahmednfwela 🎉

jonataslaw:
When I created GetNav (now replaced by this fantastic PR), my goal was to replace the current API, as it is more consistent with the navigation history and allows mock navigation.

IMO, we could migrate the current api, having a delegate by default in GetMaterialApp, and making the navigation extension methods call that delegate.

Challenges: maintain compatibility with the current api, and with packages that use GetX, such as https://pub.dev/packages/stacked_services
Or do a major release.
However, I planned to turn GetPage into a page where it could be extended, and declare the controllers, routes, guards, etc. for Getx5 on it, so I'm not sure if we can go up now, otherwise we will have to do another major release when this is ready.
IDK, that decision is up to the community.

And I would like to thank you @ahmednfwela  for the effort in making this PR, I will give you a prominent place in the changelog

ahmednfwela:
maybe we can publish this as a non-breaking change, by checking if the user provided a RouterDelegate or not, then in GetX5 we can revamp the entire api and cleanup any leftovers

jonataslaw:
Okay, I'm going to merge this now and try to do an update as soon as possible.
Thank you very much for this contribution

jasonlaw:
@ahmednfwela ,
Hi, I wish to know how to implement the login and auto login with this great PR. 
Could you please add this into your example?
Thanks!

ahmednfwela:
@jasonlaw sure, I will try to add it soon

jonataslaw:
I checked and RouteOutlet didn't have its own navigator, that was why transitions weren't working.
I think our challenge now is to create a default delegate, and call it from Get.to/Get.toNamed etc, and mainly manage that with child routes.
In the current api we use an "id" identifier to navigate nested routes. Ex:
Get.toNamed(Routes.Product, id:'foo');

I think we could create a Names:Delegates Map to do something similar with the new api. Just to remove the amount of code within the build, to be something more objective.
Ahmed, do you have any contacts?

PR #1504, opened 2021-06-02
fix flutter 2.0 error

yunchiri:
void function in Future<void> main() async{}

has Error

This expression has a type of 'void' so its value can't be used.

so fix it


jonataslaw:
Thanks, LGTM

PR #1511, opened 2021-06-03
More Improvements to the new RouterOutlet

ahmednfwela:
- Introduced `GetNavConfig` as the configuration for the router outlet (Instead of `GetPage`) which helps clear the confusion between the navigation history and the `Page` stack
- Introduced `PopMode` enum
  - `History` -> Pops the last History entry (similar to browser back button)
  - `Page` -> Pops the last `Page` in the current branch (similar to the old navigator pop)
  e.g. lets assume the user navigates directly from the `/home` route to `/home/products/1234`
  If the user goes back using `PopMode.History` , he will go back to `/home`
  But if the user goes back using `PopMode.Page`, he will go back to `/home/products/` 

- Introduced `initialRoute` in the constructor of `GetInformationParser`, to select the initial route. (instead of `Get.routeTree.routes.first`)

- Introduced `preventDuplicates` as a property to `GetPage` accompanied by the enum `PreventDuplicateHandlingMode` passed to `GetDelegate`
  ```dart
  /// Enables the user to customize the behavior when pushing multiple routes that
  /// shouldn't be duplicates
  enum PreventDuplicateHandlingMode {
    /// Removes the history entries until it reaches the old route
    PopUntilOriginalRoute,

    /// Simply don't push the new route
    DoNothing,
  }
  ```

- [Breaking change ?] Changed `RouteDecoder` to parse an entire tree now, instead of just the last page (necessary for `PopMode` to work)
  ```dart
  class RouteDecoder {
    final List<GetPage> treeBranch;
    GetPage? get route => treeBranch.isEmpty ? null : treeBranch.last;
    final Map<String, String> parameters;
    const RouteDecoder(
      this.treeBranch,
      this.parameters,
    );
  }
  ```

Also changed the example to adapt to the new changes + nested RouterOutlets

NOTE: we need to write documentation and tests for the new things, would be great if someone can help with those 😄 
@jonataslaw @Azzeccagarbugli

solves https://github.com/jonataslaw/getx/issues/1514


Azzeccagarbugli:
Nice improvements @ahmednfwela, I'm studying the code right now and trying to figure out how to setup the new pages in the documentation, I will let you know as soon as possible.

@jonataslaw are you going to release this update on the pub.dev as well?

jonataslaw:
I'll send it to pub as soon as it's stable.
I'm going to write tests for this to make sure things like arguments and parameters work in any situation.
amazing work

PR #1517, opened 2021-06-05
add StreamController bind for getx

flyingtime:
two easy way bind StreamController for getx.
1. use extension
2. use class inheritance

choice one way you like, details see getx_stream_controller_test.dart

### DIFF
--- a/lib/getx_streamcontroller.dart
+++ b/lib/getx_streamcontroller.dart
@@ -1,0 +1,24 @@
+import 'dart:async';
+import 'get.dart';
+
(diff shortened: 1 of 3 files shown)


lib/getx_streamcontroller.dart
```diff
@@ -0,0 +1,24 @@
+import 'dart:async';
+import 'get.dart';
+
+class GetxStreamController<T, S extends StreamController<T>>
```
unacorbatanegra, line 4: Why is this needed, Rx Types already do this

PR #1540, opened 2021-06-14
Refactor/rx interface notify children

kranfix:
*   Fixing linter warnings for using relative paths
*   Refactor of `notifyChilds` into `State` in favor of a unified method `RxInterface.notifyChildre` for removing duplicated code.
*   There is no need of adding test because the functionality is already tested.
*   Improving some null-safety syntax usage.


jonataslaw:
Thanks!
LGTM

PR #1545, opened 2021-06-16
Allow deleting all registered instances

lemps:
In our project we're using permanent Controllers to manage some of the applications state.

It works very well but after logout we want to reset them by reinitialisation.

Currently we have to call delete on the same ones we initialised one-by-one and this approach is prone to mistakes.

We could however use a method that deletes all the controllers from memory.

```
    Get.put(A(), permanent: true);
    Get.put(B(), permanent: true);
    Get.put(C(), permanent: true);
    
    ///
    
    Get.delete<A>(force: true);
    Get.delete<B>(force: true);
    Get.delete<C>(force: true);
```

->

```
    Get.put(A(), permanent: true);
    Get.put(B(), permanent: true);
    Get.put(C(), permanent: true);
    
    ///
    
    Get.deleteAll();
```

### DIFF
--- a/lib/get_instance/src/extension_instance.dart
+++ b/lib/get_instance/src/extension_instance.dart
@@ -109,6 +109,13 @@ extension Inst on GetInterface {
   Future<bool> delete<S>({String? tag, bool force = false}) async =>
       GetInstance().delete<S>(tag: tag, force: force);
 
+  /// Deletes all Instances, cleaning the memory and closes any open
+  /// controllers ([DisposableInterface]).
+  /// 
+  /// - [force] Will delete the Instances even if marked as [permanent].
+  Future<void> deleteAll({bool force = false}) async =>
+      GetInstance().deleteAll(force: force);
+
   void reloadAll({bool force = false}) => GetInstance().reloadAll(force: force);
 
   void reload<S>({String? tag, String? key, bool force = false}) =>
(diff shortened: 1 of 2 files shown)


lemps:
Also keep in mind that `Get.reloadAll()` nor `Get.reset()` do not reinitialise the controllers(probably did when constructor functions were supported). They also do not mark the controllers as closed as `onDelete` is never called.

jonataslaw:
Hi, thanks for this! 

You've just touched an important point for the current api:
onClose is not called in reloadAll either. Maybe, would be interesting to call onClose before cleaning up the dependency too.
In the future I will deprecate Get.reset in favor of Get.deleteAll as well, I believe this can make the api clearer.

lemps:
I fear that reload does not behave as expected with `Get.put` having the builder argument deprecated. That is because `Get.put` now expects initialised instances.

We could call the onClose methods but most likely we would need to call all the initialisation methods as well afterwards and I think the best way to do that would be to create a new instance. And that is not easy to do with builder `argument` being deprecated.

Do you feel we should manually tap into an objects life-cycle and got do reload->onClose->onInit?

lemps:
Any thoughts on this @jonataslaw ?

jonataslaw:
Well, I'll leave reloadAll for the future, for now this PR LGTM.
Thank you for your contribution

PR #1556, opened 2021-06-20
Add GetBindWidget

xdd666t:
> **I found that GetxController is automatically destroyed in the page and needs to cooperate with Getx's routing management. If you use system routing, GetxController cannot be recycled.**
>
> **example**

- jump

```dart
Navigator.push(
    Get.context,
    MaterialPageRoute(builder: (context) => SamplePage()),
);
```

- GetxController

```dart
class SampleController extends GetxController {
  var count = 0.obs;

  void increase() => ++count;
}
```

- page

```dart
class SamplePage extends StatelessWidget {
  final SampleController controller = Get.put(SampleController());

  @override
  Widget build(BuildContext context) {
    return BaseScaffold(
      appBar: AppBar(title: const Text('Simple')),
      body: Center(
        child: Obx(() {
          return Text(
            'Count ${controller.count.value} ',
            style: TextStyle(fontSize: 30.0),
          );
        }),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () => controller.increase(),
        child: Icon(Icons.add),
      ),
    );
  }
}
```

- I had to use StatefulWidget to solve this problem

```dart
class SamplePage extends StatefulWidget {
  const SamplePage({Key? key}) : super(key: key);

  @override
  _SamplePageState createState() => _SamplePageState();
}

class _SamplePageState extends State<SamplePage> {
  final SampleController controller = Get.put(SampleController());
  
  @override
  Widget build(BuildContext context) {
    return BaseScaffold(
      appBar: AppBar(title: const Text('Simple')),
      body: Center(
        child: Obx(() {
          return Text(
            'Count ${controller.count.value} ',
            style: TextStyle(fontSize: 30.0),
          );
        }),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () => controller.increase(),
        child: Icon(Icons.add),
      ),
    );
  }

  @override
  void dispose() {
    Get.delete<SampleController>();
    super.dispose();
  }
}
```

> However, using StatefulWidget makes the page look a bit complicated, so I customize **GetBindWidget**, which can easily solve this problem

- GetBindWidget

```dart
class SamplePage extends StatelessWidget {
  final SampleController controller = Get.put(SampleController());

  @override
  Widget build(BuildContext context) {
    return GetBindWidget(
      bind: controller,
      child: BaseScaffold(
        appBar: AppBar(title: const Text('Simple')),
        body: Center(
          child: Obx(() {
            return Text(
              'Count ${controller.count.value} ',
              style: TextStyle(fontSize: 30.0),
            );
          }),
        ),
        floatingActionButton: FloatingActionButton(
          onPressed: () => controller.increase(),
          child: Icon(Icons.add),
        ),
      ),
    );
  }
}
```

- this is the code of **GetBindWidget**, I hope to pass this PR

```dart
class GetBindWidget extends StatefulWidget {
  const GetBindWidget({
    Key? key,
    this.bind,
    this.tag,
    this.binds,
    this.tags,
    required this.child,
  })   : assert(
          binds == null || tags == null || binds.length == tags.length,
          'The binds and tags arrays length should be equal\n'
          'and the elements in the two arrays correspond one-to-one',
        ),
        super(key: key);

  final GetxController? bind;
  final String? tag;

  final List<GetxController>? binds;
  final List<String>? tags;

  final Widget child;

  @override
  _GetBindWidgetState createState() => _GetBindWidgetState();
}

class _GetBindWidgetState extends State<GetBindWidget> {
  @override
  Widget build(BuildContext context) {
    return widget.child;
  }

  @override
  void dispose() {
    _closeGetXController();
    _closeGetXControllers();

    super.dispose();
  }

  ///Close GetxController bound to the current page
  void _closeGetXController() {
    if (widget.bind == null) {
      return;
    }

    var key = widget.bind.runtimeType.toString() + (widget.tag ?? '');
    GetInstance().delete(key: key);
  }

  ///Batch close GetxController bound to the current page
  void _closeGetXControllers() {
    if (widget.binds == null) {
      return;
    }

    for (var i = 0; i < widget.binds!.length; i++) {
      var type = widget.binds![i].runtimeType.toString();

      if (widget.tags == null) {
        GetInstance().delete(key: type);
      } else {
        var key = type + (widget.tags?[i] ?? '');
        GetInstance().delete(key: key);
      }
    }
  }
}
```

### DIFF
--- a/lib/get_state_manager/get_state_manager.dart
+++ b/lib/get_state_manager/get_state_manager.dart
@@ -5,6 +5,7 @@ export 'src/rx_flutter/rx_getx_widget.dart';
 export 'src/rx_flutter/rx_notifier.dart';
 export 'src/rx_flutter/rx_obx_widget.dart';
 export 'src/rx_flutter/rx_ticket_provider_mixin.dart';
+export 'src/simple/get_bind_widget.dart';
 export 'src/simple/get_controllers.dart';
 export 'src/simple/get_responsive.dart';
 export 'src/simple/get_state.dart';
--- a/lib/get_state_manager/src/simple/get_bind_widget.dart
+++ b/lib/get_state_manager/src/simple/get_bind_widget.dart
@@ -1,0 +1,95 @@
+import 'package:flutter/widgets.dart';
+
+import '../../../instance_manager.dart';
+import 'get_controllers.dart';
+
+/// GetBindWidget can bind GetxController, and when the page is disposed,
+/// it can automatically destroy the bound related GetXController
+///
+///
+/// Sample:
+///
+/// class SampleController extends GetxController {
+///   final String title = 'My Awesome View';
+/// }
+///
+/// class SamplePage extends StatelessWidget {
+///   final controller = SampleController();
+///
+///   @override
+///   Widget build(BuildContext context) {
+///     return GetBindWidget(
+///       bind: controller,
+///       child: Container(),
+///     );
+///   }
+/// }
+class GetBindWidget extends StatefulWidget {
+  const GetBindWidget({
+    Key? key,
+    this.bind,
+    this.tag,
+    this.binds,
+    this.tags,
+    required this.child,
+  })   : assert(
+          binds == null || tags == null || binds.length == tags.length,
+          'The binds and tags arrays length should be equal\n'
+          'and the elements in the two arrays correspond one-to-one',
+        ),
+        super(key: key);
+
+  final GetxController? bind;
+  final String? tag;
+
+  final List<GetxController>? binds;
+  final List<String>? tags;
+
+  final Widget child;
+
+  @override
+  _GetBindWidgetState createState() => _GetBindWidgetState();
+}
+
+class _GetBindWidgetState extends State<GetBindWidget> {
+  @override
+  Widget build(BuildContext context) {
+    return widget.child;
+  }
+
+  @override
+  void dispose() {
+    _closeGetXController();
+    _closeGetXControllers();
+
+    super.dispose();
+  }
+
+  ///Close GetxController bound to the current page
+  void _closeGetXController() {
+    if (widget.bind == null) {
+      return;
+    }
+
+    var key = widget.bind.runtimeType.toString() + (widget.tag ?? '');
+    GetInstance().delete(key: key);
+  }
+
+  ///Batch close GetxController bound to the current page
+  void _closeGetXControllers() {
+    if (widget.binds == null) {
+      return;
+    }
+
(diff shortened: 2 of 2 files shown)


unacorbatanegra:
Why is this needed

CNAD666:
> If you are not using the getx route management, you can use getbuilder to create and handle the lifecycle of your controllers, this pr is trying to do the same getbuilder already fix

I know GetBuilder can solve this problem, GetBuilder can automatically remove related controllers
But if you use Obx state management, there will be a problem that the controller cannot be recycled; the PR I mentioned is for people who use Obx state management, a simpler way to solve the problem, instead of using StatefulWidget

unacorbatanegra:
What recycle means to you

CNAD666:
> What recycle means to you, I use obx, and I never have a problem

Obx recycling is carried out in the routing management of Getx
However, not everyone can change to Getx's routing management

unacorbatanegra:
The `GetBuilder` widget already handle the lifecycle of the controller, create it when the widget is on the widget tree, and delete it when it's no longer on the widget tree

CNAD666:
> The `GetBuilder` widget already handle the lifecycle of the controller, create it when the widget is on the widget tree, and delete it when it's no longer on the widget tree

Hi, buddy, we are discussing an issue of Obx state management. I don’t want to use GetBuilder at the same time when using Obx.
My PR can solve this hidden danger more simply for Obx state management. You only need to use this Widget when you need it, not necessarily

unacorbatanegra:
Your PR doesn't solve anything, 
1. Explain why this is needed.
2. What do you mean with recycle.
3. What's the "issue"

CNAD666:
> Your PR doesn't solve anything,
> 
> 1. Explain why this is needed.
> 2. What do you mean with recycle.
> 3. What's the "issue"

When I use Obx state management, without using Getx routing management, there will be a problem that the controller cannot be recycled. I can use StatefulWidget to solve it, but my Widget can be solved more simply

I think you have not finished reading the demo I wrote in PR at all, so you came to raise this series of non-objectionable questions.

Ascenio:
This looks to me similar to what [provider](https://pub.dev/packages/provider) does in the sense that we are setting scopes for dependencies. In that sense, I think it would be nice to have similar API to help, like so:

```dart
GetBindWidget(
  create: (context) => SampleController(),
  child: YourPage(),
),
```
I don't it would hurt to name it as `Provider` also because it's unlikely that people will use both provider and getx in the same project, but also it helps people who wants to migrate from one to another.

PR #1584, opened 2021-06-29
(fix) import error in example

aadarshadhakalg:
There was a import error in example, transaltion_service.dart file. This pull request fixes those errors.


roipeker (approved):
Thanks.

PR #1606, opened 2021-07-06
A RxObject should be == self

mingchaogui:
For now, a RxList is not equal to itself.

### DIFF
--- a/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
+++ b/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
@@ -82,9 +82,9 @@ mixin RxObjectMixin<T> on NotifyManager<T> {
   @override
   // ignore: avoid_equals_and_hash_code_on_mutable_classes
   bool operator ==(Object o) {
+    if (o is RxObjectMixin<T>) return value == o.value;
     // Todo, find a common implementation for the hashCode of different Types.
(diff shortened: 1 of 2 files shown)


lib/get_rx/src/rx_types/rx_core/rx_impl.dart
```diff
@@ -82,9 +82,9 @@ mixin RxObjectMixin<T> on NotifyManager<T> {
   @override
   // ignore: avoid_equals_and_hash_code_on_mutable_classes
   bool operator ==(Object o) {
+    if (o is RxObjectMixin<T>) return value == o.value;
     // Todo, find a common implementation for the hashCode of different Types.
     if (o is T) return value == o;
-    if (o is RxObjectMixin<T>) return value == o.value;
```
jonataslaw, line 87: I think this might cause problems, instead the correct syntax might be:

` if (o is RxObjectMixin<T>) return (value == o.value || this == o);`

jonataslaw (changes requested):

lib/get_rx/src/rx_types/rx_core/rx_impl.dart
mingchaogui: There is no need to compare ’this’ and ’o’. If ‘this == o’ is True, also ‘value == o.value‘.

PR #1614, opened 2021-07-08
Fix RxInterface.proxy losing its previous value on exception

WillowWisp:
Currently, whenever we encounter the `!canUpdate` exception, the static `RxInterface.proxy` isn't reverted to its previous value, making it really unsafe for other `Obx`/`Rx` (I personally encountered a problem where an `Obx` doesn't listen correctly to its `.obs`'s if there was an exception).

And yes I see that this exception should be avoided in the first place, but still, this is much safer imho.

## Pre-launch Checklist

- [-] I updated/added relevant documentation (doc comments with `///`).
- [-] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt.
- [v] All existing and new tests are passing.


jonataslaw:
Yeah, makes sense.
Thanks for this PR.
LGTM

PR #1630, opened 2021-07-14
Update README.ru.md

4uva4ok1905:
Fix error "This expression has a type of 'void' so its value can't be used"


jonataslaw:
Hi, thanks for that.
Could you let me know if other languages have the same problem? If so, could you change it to all languages?

Issue #1642, opened 2021-07-18
navigation 2.0 : Root navigator error


JustonJ99:
**Describe the bug**
@ahmednfwela 
Tested latest navigator from the example_nav2 and found Root navigator issue.
Error : Pages in a router outlet shouldn't participate in the root navigator

Bug:
1) Page navigation bar will missing for both bottom navigation bar and side menu bar

**To Reproduce**
1. Go to Products or Settings page
2. See error

Noted: 
1) Home page and profile page (before do login) without the issue.

**Screenshots**
![image](https://user-images.githubusercontent.com/37526525/126078386-97a91c86-a07d-46d1-8d09-51cac8fc6597.png)

ahmednfwela:
it's just as it says, a page shouldn't appear in a router outlet while participating in the root navigator at the same time.
Because in that case the route will be laid out twice, and it will cause duplicate key issues

ahmednfwela:
oh, it's because the commit in https://github.com/jonataslaw/getx/commit/aefbfe5514d24ceaf91a703edbb3ba91b4a33554
@jonataslaw can you revert the changes made to `participatesInRootNavigator` back? it should be null by default

JustonJ99:
@ahmednfwela Thanks for the fix. Will give the feedback after pushed.

PR #1644, opened 2021-07-19
Navigation fixes

ahmednfwela:
- rolled back `participatesInRootNavigator`
- made `GetNavigator` key optional
- Split dashboard and home routes (in example_nav2)
- Removed 'Pages in a router outlet shouldn't participate in the root navigator' restriction
- Added Navigator key to dialogs as an optional parameter 

fixes https://github.com/jonataslaw/getx/issues/1642


jonataslaw:
LGTM

PR #1647, opened 2021-07-20
Value setting is not documented properly

Gkemon:
I invest a lot of time to set value and I was a beginner in flutter so the doc was so confusing to me about how to set values

*Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots.*

## Pre-launch Checklist


jonataslaw:
Thanks. That's a good point.
I would just say to add that in case of Collections (List, Map and Set), `assignAll(newValue)` is used instead of `.value = newValue`.
Could you add this?

Issue #1650, opened 2021-07-20
How to handle async code in route guards?


MoonyDevs:
Hi,

First many thanks for the work done here and the framework.

I had a question that I posted here but I believe I should have posted it in the issues : https://github.com/jonataslaw/getx/discussions/1566

Hi all, does anybody has used https://github.com/jonataslaw/getx#redirect with async code?

For instance if authservice.isAuthenticated return a future like this:

```dart
class AuthMiddleware extends GetMiddleware {
  static const String loginRoute = "/login";
  @override
  Future<RouteSettings?> redirect(String? route) async {
    final AuthenticationApi _authService = AuthenticationApi();
    return _authService.isUserAuthenticated().then(
            (bool isAuthenticated) => 
                RouteSettings(name: isAuthenticated ? route : loginRoute) 
    );
  }
}
```

PR #1652, opened 2021-07-21
[Nav2] Improved router outlet and get delegate

ahmednfwela:
- Changed `GetMiddleware.redirectDelegate` to be async, fixes https://github.com/jonataslaw/getx/issues/1650
- Rolled back `GetRouterOutlet` default constructor, and introduced a new simpler API `GetRouterOutlet.fromRoute`
- Rolled back `PagesListExt`


steven-spiel:
I'm getting a `Null check operator used on a null value` error when using `GetRouterOutlet.builder` on this commit (bb4bf03285b700285e422c29aa70265282518567), but I'm not hitting that error on main (54e5ffec6782ad8d7a08480957bcb3faa53ca659). 

```
======== Exception caught by widgets library =======================================================
The following _CastError was thrown building Obx(has builder, dirty, state: _ObxState#a8876):
Null check operator used on a null value

The relevant error-causing widget was: 
  Obx file:///Users/stevenspiel/mobile/foo/lib/views/me/me.dart:22:12
When the exception was thrown, this was the stack: 
#0      new RouterOutlet.builder (package:get/get_navigation/src/nav2/router_outlet.dart:19:66)
#1      new GetRouterOutlet.builder (package:get/get_navigation/src/nav2/router_outlet.dart:139:14)
#2      Me.build.<anonymous closure> (package:foo/views/me/me.dart:35:31)
#3      Obx.build (package:get/get_state_manager/src/rx_flutter/rx_obx_widget.dart:70:28)
#4      RxInterface.notifyChildren (package:get/get_rx/src/rx_types/rx_core/rx_interface.dart:26:27)
```

steven-spiel:
The offending line is 

```dart
 })  : routerDelegate = delegate ?? Get.delegate<TDelegate, T>()!,
```

steven-spiel:
Fixed it by calling `Get.rootDelegate` which creates the delegate.

jonataslaw:
I have a problem here, as RouterOutlet was released in release, its api cannot be changed anymore, so RouterOutlet.fromRoute has to be a new api, and not replace the default api.
We can only make changes to the api in break versions (GetX 5).
Also, toNamed should return Future<T> instead of Future<void>, for when the user comes back and sends a result to the route that called it.
Also, just a question (I haven't really tested this, it's just a guess), if we delete setInitialRoutePath, as is being done here, when the user accesses directly from url: /home/products, will he be directed to the correct page? And when the initial route receives a pop? By the name of the function, it should be used on these occasions.
When routeGuards async, it looks amazing to me.

ahmednfwela:
@jonataslaw The problem I have with the current API, is just providing the initialRoute doesn't cover all of the possible cases for the router outlet, so at least the old one should've been left as-is, and fiddle with new constructors as we want.
The current api in itself is a bug and I don't think waiting for a major release to fix it is a good idea.

As for toNamed return type changes, I was using a completer pattern, that is, create a completer one the route is pushed, and complete it with a result when it's popped.
However, this wasn't a good pattern, and came with alot of problems when I tested it out in production apps. 
The best approach I have found for modal routes, is the `thenTo` parameter like the one I did in the nav2 example for the login screen

https://github.com/jonataslaw/getx/blob/master/example_nav2/lib/app/modules/login/views/login_view.dart#L34

this makes sure that the route is reproduceable and makes navigation more predictable than waiting for a completer to finish.

`setInitialRoutePath` by default calls `setNewRoutePath` :
```dart
  /// Called by the [Router] at startup with the structure that the
  /// [RouteInformationParser] obtained from parsing the initial route.
  ///
  /// This should configure the [RouterDelegate] so that when [build] is
  /// invoked, it will create a widget tree that matches the initial route.
  ///
  /// By default, this method forwards the [configuration] to [setNewRoutePath].
  ///
  /// Consider using a [SynchronousFuture] if the result can be computed
  /// synchronously, so that the [Router] does not need to wait for the next
  /// microtask to schedule a build.
  Future<void> setInitialRoutePath(T configuration) {
    return setNewRoutePath(configuration);
  }
```

jonataslaw:
I understand that it can't cover every case, but it does cover the main case that is used, which is BottomNavigation and NavigationRails (As in the example app, I didn't have a problem with it).
Is there a possibility to create a new builder? Or create a new class?
Your api is more configurable, and I think we should provide something like that, however, I can't change class properties because we follow SEMVER.

ahmednfwela:
the example app doesn't cover cases where the anchor route is very deep (e.g. skip 3 pages and take 1 page), which happens alot in production apps.
Also should it be a problem to break semantic versioning though? I mean the release was just yesterday, so it shouldn't be that much of a problem to change the api, especially when the current one can cause problems

jonataslaw:
Yes, I know that flexibility is required in many cases, what I'm suggesting is just add the GetRouterOutlet.fromRoute( constructor and leave the GetRouterOutlet() constructor as it is now, or add optional parameters, and follow a way when one parameter be null

ahmednfwela:
how about this, @jonataslaw it doesn't break the public api, and keeps everything working

jonataslaw:
Nice, I liked that. The only addition I would do, is if anchorRoute is null, it takes the parent route, what this code here did:

```dart
  final length = Uri.parse(initialRoute).pathSegments.length;
            return currentNavStack.currentTreeBranch
                .skip(length)
                .take(length)
                .toList();
```

ahmednfwela:
there, it should be good enough now @jonataslaw

jonataslaw:
Thanks! 
LGTM

PR #1663, opened 2021-07-22
Add scaffoldMessangerKey to GetMaterialApp class

alionour:
Recently Flutter made changes about how snackbar should appear and made snackbars to show without depending on current context as if snackbar is shown and current route is removed snackbar will still appear so in this PR and i extended GetMaterialApp with scaffoldMessangerKey.

### DIFF
--- a/example_nav2/windows/flutter/generated_plugin_registrant.cc
+++ b/example_nav2/windows/flutter/generated_plugin_registrant.cc
@@ -2,6 +2,8 @@
 //  Generated file. Do not edit.
 //
 
+// clang-format off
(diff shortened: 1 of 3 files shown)


jonataslaw:
Thank you for that. I was actually looking at it now, because I plan to remove the SnackRoute, and create a ScaffoldMessenger with our snackbar features. The advantage of this is that you can have a Snackbar on all screens, since the ScaffoldMessenger class is inside MaterialApp's WidgetApp builder.
I will work on it. Thanks.
LGTM

PR #1673, opened 2021-07-22
Added splash screen example

ahmednfwela:
- Added splash screen example
- changed `GetNavConfig.toString` to be more useful


gairick-saha (approved):

ahmednfwela:
@jonataslaw What do you think about this ?

PR #1707, opened 2021-07-29
Make `Rx<T>().stream` null-safe

steven-spiel:
RxImpl exposes the getter `stream`, which simply calls `subject.stream`, but transforms T to be optional, making these two statements incongruent:

```dart
true.obs.subject.stream.listen<bool>((bool val) => val);
true.obs.stream.listen<bool?>((bool? val) => val); // forced to be optional
```

Working on this checklist:

## Pre-launch Checklist


jonataslaw:
Thanks!
LGTM

Issue #1700, opened 2021-07-28
Get Nav2: offNamed does not remove controller


eduardoflorence:
**Describe the bug**
1. `Get.rootDelegate.offNamed` does not remove the controller from the current route from memory.
2. `initialRoute` in `GetRouterOutlet` does not put the route defined in this property in the navigation history

**Reproduction code
```dart
import 'package:flutter/material.dart';
import 'package:get/get.dart';

void main() {
  runApp(GetMaterialApp.router(
    getPages: [
      GetPage(
        name: '/',
        page: () => RootPage(),
        participatesInRootNavigator: true,
        children: [
          GetPage(
            name: '/splash',
            page: () => SplashPage(),
            binding: SplashBinding(),
          ),
          GetPage(
            name: '/login',
            page: () => LoginPage(),
            binding: LoginBinding(),
          ),
          GetPage(
            name: '/home',
            page: () => HomePage(),
            binding: HomeBinding(),
          ),
        ],
      ),
    ],
  ));
}

/* ROOT */
class RootPage extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return GetRouterOutlet.builder(
      builder: (context, delegate, current) {
        final title = current?.location;
        print(title);
        return Scaffold(
          appBar: AppBar(title: Text(title ?? '')),
          body: GetRouterOutlet(
            initialRoute: '/splash',
          ),
        );
      },
    );
  }
}

/* SPLASHSCREEN */
class SplashBinding implements Bindings {
  @override
  void dependencies() {
    Get.put(SplashController());
  }
}

class SplashController extends GetxController {
  void gotoLogin() async {
    Get.rootDelegate.offNamed('/login');
  }
}

class SplashPage extends GetView<SplashController> {
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: ElevatedButton(
          onPressed: () => controller.gotoLogin(),
          child: Text('Go Login'),
        ),
      ),
    );
  }
}

/* LOGIN */
class LoginBinding implements Bindings {
  @override
  void dependencies() {
    Get.put(LoginController());
  }
}

class LoginController extends GetxController {
  gotoHome() {
    Get.rootDelegate.offNamed('/home');
  }
}

class LoginPage extends GetView<LoginController> {
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: ElevatedButton(
          onPressed: () => controller.gotoHome(),
[...]

PR #1722, opened 2021-08-01
added PageBindings that references GetPageRoute in  [dependencies] method

ahmednfwela:
this adds `abstract class PageBindings extends Bindings` that references the `GetPageRoute` object in its `void dependencies([GetPageRoute? page]);`

useful in navigation 2.0 when using `GetRouterOutlet` with `initialRoute` that requires parameters (Since the route isn't added to the history stack, we can't access its parameters from the router delegate).

### DIFF
--- a/lib/get_instance/src/bindings_interface.dart
+++ b/lib/get_instance/src/bindings_interface.dart
@@ -1,3 +1,5 @@
+import '../../get_navigation/src/routes/default_route.dart';
+
 import 'get_instance.dart';
 
 /// [Bindings] should be extended or implemented.
@@ -8,6 +10,11 @@ import 'get_instance.dart';
 // ignore: one_member_abstracts
 abstract class Bindings {
   void dependencies();
+}
+
+abstract class PageBindings extends Bindings {
+  @override
+  void dependencies([GetPageRoute? page]);
 }
 
 /// Simplifies Bindings generation from a single callback.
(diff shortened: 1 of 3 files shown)


jonataslaw:
Nice, but I think GetRouterOutlet not being added to the stack history is the real issue here.

ahmednfwela:
it shouldn't tho, since the actual route being displayed is the initial route, and it just delegates the view part to a different route

jonataslaw:
But if we navigate with the button, the url displayed is from the initial route, that seems inconsistent to me. It's actually the only thing stopping me from migrating the current app of the company I work to nav 2, and that's sweet, because nav1 doesn't display urls correctly.

It also seems to bother other users, and cause bugs in the rest of the framework, such as the lack of removal of controllers linked to the initial route.

https://github.com/jonataslaw/getx/issues/1700

IMO this behavior is inconsistent, and unless explicitly stated, the home page should be added to history.
It also seems to bother other users, and cause bugs in the rest of the framework, such as the lack of removal of controllers linked to the initial route.

ahmednfwela:
I don't think that's a problem.

the user should have complete control over what to display based on the current navigation stack. 

the developer can even completely ignore initial route and use emptyWidget builder instead from pick pages. 

also I am using the current nav 2 system in more than 5 production apps with all sorts of cases, and I never notice any problems with disposing controllers and such.

ahmednfwela:
I also added a reply on #1700 on how to add the initial route to the stack

PR #1730, opened 2021-08-06
Fixed timeout issue

jasonlaw:
The timeout property is not being used in GetConnect.
Fixed to utilize the timeout value:
- Mobile - The value is applied to both connection and request timeout.
- Web - The value is applied to request timeout.

### DIFF
--- a/lib/get_connect/http/src/http.dart
+++ b/lib/get_connect/http/src/http.dart
@@ -205,6 +205,7 @@ class GetHttpClient {
       if (authenticate) await _modifier.authenticator!(request);
       final newRequest = await _modifier.modifyRequest<T>(request);
 
+      _httpClient.timeout = timeout;
       var response = await _httpClient.send<T>(newRequest);
 
       final newResponse =
--- a/lib/get_connect/http/src/http/html/http_request_html.dart
+++ b/lib/get_connect/http/src/http/html/http_request_html.dart
@@ -23,6 +23,9 @@ class HttpRequestImpl implements HttpRequestBase {
   ///on different sites. The default is false
   final bool withCredentials;
 
+  @override
+  Duration? timeout;
+
   /// Sends an HTTP request and asynchronously returns the response.
   @override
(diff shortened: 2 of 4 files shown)


jonataslaw:
Very nice!
LGTM

gushendra86:
Hello,

I want to ask how to get the error if the http request is timeout.

I use this code but i cannot get the result if the request is timeout :
try{
  post(url+'/student/addStudent', postData).then((value){
    print('val : '+value.statusCode.toString());
  });
} catch (err){
  print("err : "+err.toString());
}

I need to show a snackbar when the http request is timeout.

Thank you and kind regards

gushendra86:
Anyway this i think it still have a bug.

i write this code on base_provider.dart
```
import 'package:get/get_connect/connect.dart';

class BaseProvider extends GetConnect {
  @override
  void onInit() {
    httpClient.defaultContentType = "multipart/form-data";
    httpClient.timeout = Duration(seconds: 2);

    super.onInit();
  }
}
```

And then on UserProvider i write this :
```
import 'dart:async';

import 'package:myget/providers/base_provider.dart';

class UserProvider extends BaseProvider {
  final url = 'http://192.168.0.4/student/api';
  final sessionid = '5f0e6bfbafe255.00218389';

  // add user
  Future<Response> postData(String name, String email, String phone) async {
    final postData = FormData({
      'session_id': sessionid,
      'student_name': name,
      'student_phone_number': phone,
      'student_gender': 'male',
      'student_address': email
    });

    var response = await post(url+'/student/addStudent', postData);

    print(response.body.toString());

    return response;
  }
}
```

There are 2 error :

1. On my PHP API file, i add code sleep(4); So the API will be called after 4 second, but in the application i set the timeout is 2 second and it still run and timeout is not affect.
2. Another error is when i don't set the timeout from the onInit(). And then on the PHP API file, i add code sleep(7); After that when i call function postData, there are no result. This error is not happen when i use get 4.1.4, but the error happen at get 4.3.8

Thank you and kind regards

PR #1747, opened 2021-08-09
Allow GetRouterOutlet to specify GetDelegate

steven-spiel:
GetRouterOutlet is being forced to use the rootDelegate, but there are instances where we may need to specify the delegate, like when multiple delegates may be needed. This simple change provides that ability.

## Pre-launch Checklist


jonataslaw:
Nice! 
LGTM

PR #1766, opened 2021-08-13
fix try/catch case missed in socket_notifier

ShookLyngs:
When JSON content is a List type, the catch syntax with `on Exception` in `_tryOn()` won't catch the error.
Deleting `on Exception` will the bug.

## Pre-launch Checklist


jonataslaw:
Hi, Thanks for your contribution

PR #1815, opened 2021-08-28
Missing arguments of copyWith

rodrigorahman:
Some attributes were being overwritten, such as bodyBytes, causing a problem in their copy inside the addRequestModifier method.


jonataslaw:
Thanks! 
LGTM

PR #1828, opened 2021-09-02
Fix RxList.removeWhere calls 'ever' multiple times

zuvola:
A single call to RxList.removeWhere will call "ever" multiple times.
The same goes for retainWhere.

It should pass the following test, but in fact it does not.

```dart
    final list = [1, 2, 3].obs;
    var count = 0;
    ever<List<int>>(list, (value) {
      count++;
    });
    list.removeWhere((element) => element == 2);
    await Future.delayed(Duration.zero);
    expect(count, 1);
```

So I fixed it.

## Pre-launch Checklist


jonataslaw:
Thanks! 
LGTM

PR #1841, opened 2021-09-08
.obx extension on flutter widgets

maxzod:
## Pre-launch Checklist

# .obx
the original way
```dart
Obx (() => Text(controller.name));
```
Or you can use `.obx` extension on any `Widget`
```dart
Text(controller.name).obx
```

### DIFF
--- a/documentation/en_US/state_management.md
+++ b/documentation/en_US/state_management.md
@@ -1,5 +1,5 @@
-* [State Management](#state-management)
-  + [Reactive State Manager](#reactive-state-manager)
+- [State Management](#state-management)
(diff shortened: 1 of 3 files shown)


jonataslaw:
This is something I've thought about in the past, and I gave up because I remembered that dart closes its parameters as soon as you close the constructor, that would make the reactive variables not able to be linked to your widgets, so the widget goes inside a function in Obx, because with the function the constructor is not called.
If this really works, it's going to be awesome, but I'm really intrigued how it could work since Text(rxVariable).obx would read rxVariable before Text is inserted into an Rx in extensions.

maxzod:
turns out you are right ! 😅

PR #1847, opened 2021-09-10
markAsDirty must use permenent property

zenalex:
markAsDirty method doesn't take into account permanent property of controller. So, permanent controllers are re-creating every Get.find call.

In previous versions all worked fine. I added this check
 if (dep != null && !dep.permanent) {


zenalex:
Really I see one variant. Do not use permanent in binding. But Why???

jonataslaw:
Good catch! Permanent controllers should not be marked as dirty.
Thank you so much for this!

PR #1858, opened 2021-09-12
enable logs on debug instead of always

maxzod:
this PR will make logs enabled by default in debug mode only 
in the past it was always enabled
also user is able to turn it on or Off just the default value has been changed


jonataslaw:
Hi, I found this really useful, I just have a suggestion to make.
I generally debug apps using Profile Mode, and logs are useful there.
Instead of using kDebugMode, I would like the logs to be displayed whenever: `!kReleaseMode.`

matthewkooshad:
if my launch.json profile has "flutterMode": "debug", should i still be seeing the `[GETX] Instance "..." has been created` messages?

PR #1928, opened 2021-10-21
Fix GetController to GetxController

Hanggi:


README.md
```diff
@@ -993,7 +993,7 @@ I love this Widget, is so simple, yet, so useful!
 Is a `const Stateless` Widget that has a getter `controller` for a registered `Controller`, that's all.
 
 ```dart
- class AwesomeController extends GetController {
+ class AwesomeController extends GextController {
```
Ascenio, line 996: Should be `GetxController`.

Ascenio (changes requested):

Issue #1933, opened 2021-10-24
Rx.trigger fires twice !?


gslender:
**Describe the bug**
Rx.trigger fires event to listeners twice !?

**Reproduction code

```dart
import 'package:flutter/material.dart';
import 'package:get/get.dart';

void main() {
  Get.put(MyController());
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({Key? key}) : super(key: key);
  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  final MyController controller = Get.find<MyController>();

  var counter = 0;

  @override
  void initState() {
    super.initState();

    controller.myClass.listen((myClass) {
      counter++;
      debugPrint(">>> $counter myClass $myClass ${DateTime.now().toString()}");
    });
  }

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(
          title: const Text("testGetx - trigger"),
        ),
        body: Center(
          child: Column(
            mainAxisAlignment: MainAxisAlignment.center,
            children: [
              OutlinedButton(
                onPressed: () {
                  controller.myClass
                      .trigger(MyClass(DateTime.now().toString()));
                },
                child: const Text('Click Me'),
              ),
            ],
          ),
        ),
      ),
    );
  }
}

class MyController extends GetxController {
  var myClass = MyClass.none().obs;
}

class MyClass {
  String id;

  MyClass(this.id);

  MyClass.none() : id = "none";

  @override
  String toString() {
    return id;
  }
}
```

**To Reproduce**
Click the button more than once and you will see myClass being sent as an event to the listener twice... timestamps are the same for the object being created, but the event is received twice only ms apart !?

**Expected behavior**
Only 1x trigger() results in 1x listener event occuring

**Flutter Version:**
2.5.3

**Getx Version:**
4.3.8

PR #1950, opened 2021-11-01
Fix to rx.trigger so that it checks for previous send to stream by set value

gslender:
This fixes issue #1933 

- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making or feature I am adding, or @jonataslaw said the PR is test-exempt.
- [X] All existing and new tests are passing.

### DIFF
--- a/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
+++ b/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
@@ -67,6 +67,7 @@ mixin RxObjectMixin<T> on NotifyManager<T> {
   }
 
   bool firstRebuild = true;
+  bool sentToStream = false;
 
   /// Same as `toString()` but using a getter.
   String get string => value.toString();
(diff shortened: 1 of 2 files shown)


jonataslaw:
Hi, thanks for your contribution,
I was able to reproduce this problem with your code.
Could you please duplicate this test and add different numbers?
I think this will cover us with the hypothesis of equal values and different values, which was just what we needed to identify this bug.

Sorry for the delay, we're busy with the new version and selecting only PRs that maintain backward compatibility to merge, which seems to be the case for this one.

gslender:
I'm unsure in what you're saying you've found, as the test changes I submitted in my PR have exactly different numbers being tested and the results are correct.

e.g....

```
    // we call 3
    reactiveInteger.trigger(3);
    // then repeat twice
    reactiveInteger.trigger(3);
    reactiveInteger.trigger(1);

    await Future.delayed(Duration(milliseconds: 100));
    expect(3, timesCalled);

```

gslender:
Perhaps the comments in the test are not 100% accurate, but I changed as little as possible assuming you'd want to verify and own the changes to confirm either a) that indeed my fix is correct and b) that the tests are more broadly correct

jonataslaw:
Nvm I do it

zenas:
> Nvm 我做到了

According to the difference between trigger and call, maybe we just need to consider _value == v 。

like:
`
void trigger(T v){
  if(_value == v){
    var firstRebuild = this.firstRebuild;
    if(fistRebuild){
    subject.add(v)
  }
  }else{
    value = v
  }
}
`

Issue #1975, opened 2021-11-11
Bug: Getx does not dispose controller put in the Widget "home" of GetMaterialApp


henriqueArrazao:
**ATTENTION: DO NOT USE THIS FIELD TO ASK SUPPORT QUESTIONS. USE THE PLATFORM CHANNELS FOR THIS. THIS SPACE IS DEDICATED ONLY FOR BUGS DESCRIPTION.**
**Fill in the template. Issues that do not respect the model will be closed.**

**Describe the bug**
Getx does not dispose controller put in the first widget even using "Get.offAll".

**Reproduction code
NOTE: THIS IS MANDATORY, IF YOUR ISSUE DOES NOT CONTAIN IT, IT WILL BE CLOSED PRELIMINARY)**

example:

```dart
import 'package:flutter/material.dart';
import 'package:get/get.dart';

void main() {
  runApp(const MyApp());
}

class HomeCounterController extends GetxController {
  final count = 0.obs;
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return GetMaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: HomeCounterPage(),
    );
  }
}

class HomeCounterPage extends StatelessWidget {
  final controller = Get.put(HomeCounterController());

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const Text(
              'You have pushed the button this many times:',
            ),
            Obx(
              () => Text(
                '${controller.count}',
                style: Theme.of(context).textTheme.headline4,
              ),
            ),
            ElevatedButton(
                onPressed: () {
                  Get.offAll(() => SecondScreen());
                },
                child: const Text("Next screen")),
          ],
        ),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () => controller.count.value++,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ), // This trailing comma makes auto-formatting nicer for build methods.
    );
  }
}

class SecondScreen extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: ElevatedButton(
          onPressed: () {
[...]

PR #1979, opened 2021-11-12
Fix home controller not being disposed

henriqueArrazao:
This PR fixes: https://github.com/jonataslaw/getx/issues/1975


jonataslaw:
Wow, great catch. Could you also include CupertinoPageRoute here? When using GetCupertinoApp instead of GetMaterialApp the home is a cupertino route

PR #2008, opened 2021-11-25
prepare snackbar api to getx 5

jonataslaw:
To have a context-free, page-agnostic snackbar, we used OverlayRoute to display a partial route.
However this had several problems:

**1: There was no possibility to close the page without closing the snackbar
2: Get.back() could cause problems with tests of `Get.isSnackbarOpen` not being properly invoked
3: Sometimes when using iOS popGesture with an open snackbar, some visual inconsistency might appear.
4: When going to another route, the snackbar was not displayed on the new page, and if the user clicked on the new route as soon as he received a Snackbar, he could not read it.**

This PR creates a new api, from scratch, using `Overlay` instead of `Routes` to display the new snackbar.

This api will be released as a break update because we are making deep changes to the api like:

`Get.closeCurrentSnackbar();`
`Get.closeAllSnackbars();`

Perhaps the snackbar opening method returns its controller so that it is possible to close a specific snackbar. This is a feature that doesn't exist natively in Flutter yet, and I'm looking to add it to Getx.

The dialog and bottomsheet api will probably change too. The community will be consulted about changing the api of `routes > overlay` in `dialogs` and `bottomsheets`.
This would allow using `Get.closeDialog()` and `Get.closeBottomSheet()` instead of Get.back().

Using `Get.back()` is handy, it's a small command, which closes routes, dialogs, snackbars, bottomsheets, etc, however Getx 5 will prioritize **code safety**, and splitting will reduce the check code as well. Currently we have to check if a **snackbar is open**, to close the snackbar and prevent the app from going **back a page**, all this boilerplate code will be removed, at the cost of having what it closes in front of Get.back command.

A second proposal for this api is to continue with Get.back(), and add super parameters to it, like:

`Get.back(mode: CloseMode.routeOnly)  // default`
`Get.back(mode: CloseMode.allSnackbars)`
`Get.back(mode: CloseMode.currentSnackbar)`

And by default `Get.back()` will close everything and go back to the previous route.
I'm going to open a poll on Discord about that.

## Pre-launch Checklist


jahnli:
When will 5.0 be released

PR #2040, opened 2021-12-09
Update internacionalization.dart Fixes #2039

### LINKED ISSUE #2039: Localization don't work for Locale('en_US') or Locale('de_DE')
In the previous version of 4.51 version, Localization was working in these cases:

Locale('en_US')
Locale('de_DE')

Translation Keys:
Map<String, Map<String, String>> get keys => {
        'en_US': en_US,
        'de_DE': de_DE,
      };
      
      
But now it can't translate.

codercengiz:
Fixes #2039 
Localization don't work for Locale('en_US') or Locale('de_DE')

With this change, it supports both situations:
When locale has only language code like Locale('en') or Locale('en_US')

## Pre-launch Checklist


jonataslaw:
LGTM

PR #2063, opened 2021-12-18
DOCS: minor grammar fix

connected-tcaceres:
Very minor one line grammar fix.

## Pre-launch Checklist
Test exempt, documentation update.


connected-tcaceres:
Hello @jonataslaw , is it possible to merge this one liner change? 🙏

README.md
```diff
@@ -141,7 +141,7 @@ void main() => runApp(GetMaterialApp(home: Home()));
 ```
 
 - Note: this does not modify the MaterialApp of the Flutter, GetMaterialApp is not a modified MaterialApp, it is just a pre-configured Widget, which has the default MaterialApp as a child. You can configure this manually, but it is definitely not necessary. GetMaterialApp will create routes, inject them, inject translations, inject everything you need for route navigation. If you use Get only for state management or dependency management, it is not necessary to use GetMaterialApp. GetMaterialApp is necessary for routes, snackbars, internationalization, bottomSheets, dialogs, and high-level apis related to routes and absence of context.
-- Note²: This step is only necessary if you gonna use route management (`Get.to()`, `Get.back()` and so on). If you not gonna use it then it is not necessary to do step 1
+- Note²: This step is only necessary if you are going to use route management (`Get.to()`, `Get.back()` and so on). If you are not going to use it then it is not necessary to do step 1
```
Ascenio, line 144: I fell like an `otherwise` could be a good fit to simplify the second phrase.

PR #2086, opened 2021-12-27
Improvement on RxSet: don't refresh() when adding an already present value

Marplex:
The reactive Set implementation should not update the Stream when adding an already present value. Since the Set remains the same, no new changes should be notified.

The change is very minimal and only affects the `get_rx` library, with no impact on any tests.

## Pre-launch Checklist


jonataslaw:
This makes perfect sense, thanks for your contribution.
I will merge soon.

PR #2094, opened 2022-01-02
Personalizable Border Radius in Default Dialog

jiurchuk:
Border Radius in Default Dialog defaults to 100 and radius received via param was not used at all. 
This PR changes that: If user specifies a value, then that value is used. Otherwise, 100 will be used as always.

## Pre-launch Checklist


jonataslaw:
Thanks! LGTM

PR #2142, opened 2022-01-20
Update rx_notifier.dart

apieceofcode1801:
fix refresh() not called when status equals to SuccessState


jonataslaw:
I don't know how I missed entering this, 
Thank you!
LGTM

PR #2156, opened 2022-01-28
Null safe Get.engine

CaLouro:
I have made the `Get.engine` getter null safe to facilitate its use. Plus, I've removed the null check operator `!` used within the package that was always used after the getter call.

The updated function looks like this:
```
  /// The current null safe [WidgetsBinding]
  WidgetsBinding get engine {
    return WidgetsFlutterBinding.ensureInitialized();
  }
```

Previously it looked like this:
```
  ///The current [WidgetsBinding]
  WidgetsBinding? get engine {
    if (WidgetsBinding.instance == null) {
      WidgetsFlutterBinding();
    }
    return WidgetsBinding.instance;
  }
```
which was just a non null safe version of the `WidgetsFlutterBinding.ensureInitialized()` function, that's why the new version is just a call to that.

Note that this change will throw warnings on IDEs and during compilation, but it should NOT break code or compilation. 

Let met know if I need anything more on this PR.

## Pre-launch Checklist

### DIFF
--- a/lib/get_navigation/src/extension_navigation.dart
+++ b/lib/get_navigation/src/extension_navigation.dart
@@ -1059,7 +1059,7 @@ extension GetNavigationExt on GetInterface {
   /// Your entire application will be rebuilt, and touch events will not
   /// work until the end of rendering.
   Future<void> forceAppUpdate() async {
-    await engine!.performReassemble();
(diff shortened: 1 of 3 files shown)


jonataslaw:
It's makes sense, thank you!

PR #2181, opened 2022-02-10
fix: fix Chinese translation format of state_management.md

KouYiGuo:
**modified file**

documentation/zh_CN/state_management.md

**reason**

When markdown source file contains Chinese characters, especially Chinese character lying with markdown keyword, like "_" and "__", the parser can not recognize the keyword. So we must type space between the Chinese character and the keyword.

For example, `你好_hello_` will be rendered as "你好_hello_" , and `你好 _hello_ ` will be rendered as "你好 _hello_" 。


jonataslaw:
Thank you! 
LGTM

PR #2207, opened 2022-02-22
defaultDialog should allow for no title

banderberg:


banderberg:
@jonataslaw thanks for the suggestion.

PR #2309, opened 2022-04-14
Replace class declaratrion with typedef

Andrflor:
From dart 2.13 typedef is now a thing.
Using class extensions for declaring RxBool, RxString.... is breaking generics.
Also boilerplate code is reduced with no more need to create class specific obs extension.

The main idea is to make RxBool, RxString... what they should be, type aliases for Rx<bool> Rx<String>...

This make pretty convenient functions easy to write without having to care about any specificity what so ever.
For example i included also a pipe function that allow to pipe one observable into another.


jonataslaw:
1- I agree with this, and I think this is better than using inheritance (current approach).

2- However, some Rx have different behavior, we can use Extensions for that. For example, the toString() method of an RxBool, will print true/false instead of printing (instance of RxBool).

We can add typedef to types, and add content inside extensions. I'm not sure if this is possible, but it might be an alternative.

Andrflor:
If you are not overriding stuff it's not a problem.
The only override so far is the toString method.
(Rx iterables are out of scope).

For this toString case if you want the value to be printed why not just declare?

```dart
class Rx<T> extends _RxImpl<T> {
  Rx(T initial) : super(initial);

  @override
  dynamic toJson() {
    try {
      return (value as dynamic)?.toJson();
    } on Exception catch (_) {
      throw '$T has not method [toJson]';
    }
  }

  @override
  String toString() => value.toString();
}
```

Or if only for booleans why not using type asset?
```dart
class Rx<T> extends _RxImpl<T> {
  Rx(T initial) : super(initial);

  @override
  dynamic toJson() {
    try {
      return (value as dynamic)?.toJson();
    } on Exception catch (_) {
      throw '$T has not method [toJson]';
    }
  }

  @override
  String toString() =>
      value.runtimeType == bool ? value.toString() : super.toString();
}
```

PR #2350, opened 2022-05-12
I removed "!" from the instance property

amrhassann:
that because since the Flutter 3.00 the instance getter will not return null anymore. As you can see here: binding.dart#L304 The commit associated with this change is this one: ab89ce285f76a3fd3a9328ec863f5623cc5ac8bc


HamadaAllipy:
great job , it's working for me  💚😁

PR #2403, opened 2022-06-06
Added README-hi.md for README in Hindi Language.

Adison-Masih:
It is the README.md in Hindi Language Mostly Spoken In Asian Countries, Especially India. It Currently Covers Topics Till Route Management But Would Will Gradually  Translate It Completely.


jonataslaw:
Thank you so much for this.
It would be great to have the specific files for state management, dependency injection and routes translated as well, but the main document is already a huge improvement.

PR #2421, opened 2022-06-20
Fixed bug in dependency deletion.

MaximumCookie:
Before the Binding creates dependencies for a new route, the dependencies from the old route are marked as "dirty".  However, if the new route adds one of those same dependencies, then the dirty dependency is moved to the 'lateRemove' property and the root dependency is no longer marked 'isDirty'.  This means, when Get.delete() is eventually called on the dependency, the lifecycle 'delete()' is called on the new dependency instead of the old one, and the old dependency ends up being removed from memory (To summarize: The delete() method is called on the wrong controller if both the old and new route have the same controller in their Binding.).

It seems like the delete() lifecycle method should always be called on the same dependency that ends up being removed from memory, which this change fixes.

## Pre-launch Checklist

### DIFF
--- a/lib/get_instance/src/get_instance.dart
+++ b/lib/get_instance/src/get_instance.dart
@@ -354,7 +354,7 @@ class GetInstance {
     if (dep == null) return false;
 
     final _InstanceBuilderFactory builder;
-    if (dep.isDirty) {
+    if (dep.lateRemove != null) {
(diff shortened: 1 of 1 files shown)


jonataslaw:
Wow, thanks for that. Looks like you're right.
As this is critical, I'll write some tests and see if it doesn't cause any side effects and then I merge it.

PR #2520, opened 2022-08-26
Fixed generic type and equality assert

Andrflor:
1 - The generic parameters were not properly passed to the factories
2 - The equality was wrong when same prop but different class types

For example:
```dart
EmptyStatus()==LoadingStatus() // return true
```


jonataslaw:
Thanks for your contribution! 
LGTM

PR #2585, opened 2022-10-21
Support TLDs of up to 7 characters

alindeman:
`isURL` only supports TLDs up to 6 characters long. However, there are several common TLDs with 7 or more letters: `.example`, `.systems`, etc.

I would be open to increasing this number if desired, because according to RFC 1034, the TLD could be as long as 63 characters.

## Pre-launch Checklist


jonataslaw:
Thanks for your contribution! 
LGTM

PR #2590, opened 2022-10-24
Changed the 'dialogue made in 3 lines' part.

Ahmadexe:
The dialogue box, as coded in extension_navigation.dart shows an unprofessional text, "Dialogue made in 3 lines code.", which looks poor in production and to alter the middletext manually using sizedbox shrink is pointless.

Hence, I changed it to, "\n" which just adds an empty line instead of the "Dialogue made in 3 lines" thing.


jonataslaw:
I strongly agree with you. 
Thanks for your contribution! 
LGTM

PR #2620, opened 2022-11-17
Theme pageTransition

Khurshidddbek:
Currently, GetX does not use theme pagaTransition, default transition is not taken from theme. It is somehow hard to customize pageTranstion though CustomTransition. That’s why I changed default page transtion.
It is easy to set custom transition through ThemeData.


jonataslaw:
Fair enough. 
LGTM, thanks!

Issue #2480, opened 2022-08-01
leak fix


qyxghcl:
class GetPageRoute<T> extends PageRoute<T>
    with GetPageRouteTransitionMixin<T>, PageRouteReportMixin {

   @override
   void dispose() {
     super.dispose();
     final middlewareRunner = MiddlewareRunner(middlewares);
     middlewareRunner.runOnPageDispose();
     /// there need to release the _child
     _child = null;
   }
}

huangxc212:
yes, Leakage caused by variable _child!

![leak](https://user-images.githubusercontent.com/37100355/202386113-8e49ecc0-ac92-4eeb-9d3f-228bb9bb2a13.jpg)

jonataslaw:
Could you make a PR with this change, please?

huangxc212:
I make a PR,please check,thanks!
buy the way, I found an other leakage in Class RouterReportManager,ref #1854
mybe remove the key when value is empty?
![20221125111425](https://user-images.githubusercontent.com/37100355/203942924-0a92d440-ef2f-4fc9-bf5e-c1cb5839b667.jpg)
![20221125111618](https://user-images.githubusercontent.com/37100355/203942958-e5955e39-913b-42e1-95ef-2440283488fc.jpg)

Issue #1854, opened 2021-09-12
RouterReportManager  memory leak


wangronghui01:
Now we find that Getx has a memory leak problem. The RouterReportManager uses global when managing dependencies_ Routeskey: when a route is obtained by get.find, a value will be inserted into the map. The key of the map uses the getpageroute object. However, getpageroute will be re created every time the re route enters, so the map.containskey comparison must be inconsistent, because the getpageroute object does not find any place to rewrite hashcode or equals. I don't know if it's my usage.

example:

```dart
static void appendRouteByCreate(GetLifeCycleBase i) {
    _routesByCreate[_current] ??= HashSet<Function>();
    // _routesByCreate[Get.reference]!.add(i.onDelete as Function);
    _routesByCreate[_current]!.add(i.onDelete);
  }

  static void reportRouteDispose(Route disposed) {
    if (Get.smartManagement != SmartManagement.onlyBuilder) {
      WidgetsBinding.instance!.addPostFrameCallback((_) {
        _removeDependencyByRoute(disposed);
      });
    }
  }

  static void reportRouteWillDispose(Route disposed) {
    final keysToRemove = <String>[];

    _routesKey[disposed]?.forEach(keysToRemove.add);

    /// Removes `Get.create()` instances registered in `routeName`.
    if (_routesByCreate.containsKey(disposed)) {
      for (final onClose in _routesByCreate[disposed]!) {
        // assure the [DisposableInterface] instance holding a reference
        // to onClose() wasn't disposed.
        onClose();
      }
      _routesByCreate[disposed]!.clear();
      _routesByCreate.remove(disposed);
    }

    for (final element in keysToRemove) {
      GetInstance().markAsDirty(key: element);

      //_routesKey.remove(element);
    }

    keysToRemove.clear();
  }
```

**Flutter Version:**
flutter 2.5.0

**Getx Version:**
Getx ^4.3.8

jonataslaw:
In fact the route is added in Route.install, and it is removed in Route.dispose.
I can't envision a situation where Route could change its object without calling dispose.

wangronghui01:
`static void _removeDependencyByRoute(Route routeName) {
    final keysToRemove = <String>[];

    _routesKey[routeName]?.forEach(keysToRemove.add);

    /// Removes `Get.create()` instances registered in `routeName`.
    if (_routesByCreate.containsKey(routeName)) {
      for (final onClose in _routesByCreate[routeName]!) {
[...]

PR #2629, opened 2022-12-01
fix leakage in routerReportManager

huangxc212:
the key  in _ Routeskey  may remove when value is empty #1854 and  #2480


jonataslaw:
It Will investigate why it is failing in tests

PR #2680, opened 2023-02-02
Add a feature when Hovering on web for Snackbar

RafaRuiz:
I've found a few use cases for snackbars where we explain a certain error to the user, even to copy/paste some code in it.
We've seen that in the user experience that users tend to put the mouse on the top to "hold what they're reading", but snackbars are fading out anyway. Some other people aren't interested so it's not that we can keep a snackbar forever there.

By making the methods public, the ones I changed from private to public, we allow the snackbar to stop and fade out when the hover enters and exits.

## Pre-launch Checklist

- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] All existing and new tests are passing.


RafaRuiz:
I don't understand why tests are failing, but they're failing in master too:
image

will try to retry them

jonataslaw:
All tests passed. Thanks for your contribution! 
LGTM

RafaRuiz:
Thanks for your dedication @jonataslaw

PR #2815, opened 2023-06-03
Update http_request_html.dart

wheeOs:
emphasizing that response interception has still to be implemented for web


jonataslaw:
Thank you, I will implement it

PR #2832, opened 2023-06-30
Update connect.dart with null checker exception

DaZealous:
*I updated connect.dart file to remove nullable exception when mutation function is checking for "errors"*


jonataslaw:
Thank you for your contribution

PR #2837, opened 2023-07-03
Added New method Get.findOrNull ()

aryan-more:
Implemented findOrNull method

This pull request adds a new method called findOrNull to GetX. The method is similar to the existing find method but instead of throwing an error when the specified dependency is not found, it returns null.

This methods aims to avoid writing more boilerplate code like this

```
if (Get.isRegistered<ControllerX>()){
  Get.find<ControllerX>().refresh();
}
```
To this

```
Get.findOrNull<ControllerX>()?.refresh();
```

## Pre-launch Checklist

- [ Done] I updated/added relevant documentation (doc comments with `///`).
- [ Done] I added new tests to check the change I am making or feature I am adding, or @jonataslaw said the PR is test-exempt.
- [Partial] All existing and new tests are passing. There is one test in example failing but it was also failing before making changes

### DIFF
--- a/lib/get_instance/src/extension_instance.dart
+++ b/lib/get_instance/src/extension_instance.dart
@@ -298,6 +298,17 @@ extension Inst on GetInterface {
     }
   }
 
+  
+  /// The findOrNull method will return the instance if it is registered;
+  /// otherwise, it will return null.
+  S? findOrNull<S>({String? tag}) {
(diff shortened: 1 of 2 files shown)


aryan-more:
@jonataslaw

jonataslaw:
LGTM

PR #2848, opened 2023-07-15
fix: respect query and required parameters

tshedor:
The router doesn't include optional query-string parameters when required parameters are present.

For example, `/route/:id` and `/route?id=1` are respected, but `/route/:id/profile?title=dev` only parses `:id`. This PR provides `:id` and `title` to `Get.parameters`

## Pre-launch Checklist


jonataslaw:
LGTM

Issue #1560, opened 2021-06-21
transition: Transition.downToUp have a black background container


fabianogadenz:
Hello everyone, i have a problem in navigator with getx. I want use transition: Transition.downToUp (like a bottom sheet) but i have a black container in the transition'

example:

```dart
GetPage(
        name: SIMULATOR_ROUTE,
        page: () {
          return MyPage(
            preSelectedBeneficiaryId: args?.beneficiaryId,
            transactionPresenter: GetIt.I<TransactionPresenter>(),
          );
        },
        transition: Transition.downToUp,
        transitionDuration: Duration(seconds: 3)
),
```

**Expected behavior**
I would like to have a navigation without this black container.

**Screenshots*

https://user-images.githubusercontent.com/12851596/122818739-f1a27e00-d2af-11eb-91d4-cf179e608fb7.mp4

**Flutter Version:**
2.2.1

**Getx Version:**
^4.1.4

**Describe on which device you found the bug:**
ex: Iphone 11

BrandCooper:
same problem with me, is there a more elegant solution? @zeucxb  @jonataslaw

DSPerson:
> popGesture

you save me

tgdong2296:
Using `fullscreenDialog: true` and `popGesture: false` works for me, thanks a lot!

PR #2864, opened 2023-08-07
bug fixed: Transition.downToUp have a black background container #1560

Leobuaa:
- *I add a parameter ```oldPage``` on api ```GetNavigationExt.to()```, ```oldPage```  is the previous page when you push a new page, if you use a custom ```transition```, we need to change the old page exit transition, otherwise it will show black container in some cases, it fixes the bug:*
https://github.com/jonataslaw/getx/issues/1560 

- *When we use ```Get.to(() => NextPage(), transition: Transition.downToUp)```, we should change the previous page's exit transition to SlideTopTransition, so change to usage to new api: ``` Get.to(() => NextPage(), transition: Transition.downToUp, oldPage: currentPage)```*

before:

https://github.com/jonataslaw/getx/assets/6970283/a0d37c5e-387f-4703-bdea-2cf364fe5be0

after:

https://github.com/jonataslaw/getx/assets/6970283/56e5aaf4-b436-4a26-8f22-df11bb0303d4

## Pre-launch Checklist


Leobuaa:
@jonataslaw

jonataslaw:
Thank you very much for this PR!
I'm just wondering if there's a way to do this dynamically, without having to pass a page by parameter.

In fact, I think this problem only occurs on iOS, due to the cupertino animation, but this can be disabled.

If it only happens on iOS, we could disable the previous page animation when the animation is vertical.

PR #2902, opened 2023-09-05
'useInheritedMediaQuery' is deprecated and shouldn't be used

Aniketkhote:
'useInheritedMediaQuery' is deprecated and shouldn't be used. Remove this parameter as it is now ignored. MaterialApp never introduces its own MediaQuery; the View widget takes care of that. This feature was deprecated after v3.7.0-29.0.pre.


jonataslaw (approved):

jonataslaw:
Wow, I just thought of making this change and your PR was already here.

Thank you very much for your contribution!

PR #2953, opened 2023-10-30
Make cancelAllSnackbars async

seungsuyoo:
**Title:** Make `cancelAllSnackbars` async

**Description:**

This PR changes the `cancelAllSnackbars` method in `SnackbarController` to be `async`. This allows the method to wait until all snackbars are cancelled before returning. Previously, the method returned immediately, and any errors that occurred during the cancellation of the snackbars were not caught by a `try-catch` block.

This change will help prevent `LateInitializationError` errors when trying to cancel all snackbars.


jonataslaw:

PR #2957, opened 2023-11-02
Refactor code

Aniketkhote:
Refactor lots of code with used final keywords, type annotations.


lib/get_animations/animations.dart
```diff
@@ -35,37 +29,25 @@ class FadeOutAnimation extends OpacityAnimation {
...
   OpacityAnimation({
-    super.key,
-    required super.duration,
-    required super.delay,
-    required super.child,
-    required super.onComplete,
-    required double begin,
-    required double end,
-    required super.idleValue,
+    required super.duration, required super.delay, required super.child, required super.onComplete, required final double begin, required final double end, required super.idleValue, super.key,
```
yang-lile, line 32: why to do this? it is not like dart lints.

PR #3011, opened 2024-01-10
The default behavior for Get.defaultDialog's onCancel method should be closeAllDialogs();

leeyisoft:
The default behavior for Get.defaultDialog's onCancel method should be closeAllDialogs();


jonataslaw:
I agree. I'm just going to add a TODO here, because I want to change the dialogs and bottomsheets api to something more similar to snackbars, so that I can close a specific dialog.

PR #3041, opened 2024-02-28
Thai translation :)

whyzotee:
Thai docs not finish yet.


knottx:
I noticed that the document translation to Thai has some issues. It seems that it might have been translated using Google Translate, resulting in confusing and inaccurate Thai text. For better clarity and accuracy, I recommend reviewing the translation manually

However, I know those changes will be appreciated.
I suggest that we keep the document in English. Learning English is valuable not just for understanding this package, but for career growth and accessing a broader range of resources.

whyzotee:
Thank for comment 😃

PR #3047, opened 2024-03-05
fixed types in equality operator

ardakalayci:
When hashcode is not used, we can only be sure about whether primitive types have changed or not. However, for lists and maps, we need to check with hashcode.

### DIFF
--- a/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
+++ b/lib/get_rx/src/rx_types/rx_core/rx_impl.dart
@@ -100,7 +100,7 @@ mixin RxObjectMixin<T> on GetListenable<T> {
   set value(T val) {
     if (isDisposed) return;
     sentToStream = false;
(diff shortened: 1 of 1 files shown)


jonataslaw:
Can you explain this to me better?
The equality operator by default compares the hashCode of two objects.
A list will only be equal to another list if both are const. This doesn't seem to have anything to do with the hashcode.
Only immutable lists are the same as immutable lists.

```dart
void main() {
  
  List foo = [1,2,4, "eee"];
  List bar = [1,2,4, "eee"];
  
  print(foo == bar); // false
  print(foo.hashCode == bar.hashCode); // false
}
```

```dart
void main() {
  
  List foo = const  [1,2,4, "eee"];
  List bar = const [1,2,4, "eee"];
  
  print(foo == bar); // true
  print(foo.hashCode == bar.hashCode); // true
}
```

PR #3050, opened 2024-03-08
Snackbar fix

jonataslaw:
Fix Snackbar memory leak
Fix Snackbar blocking widget below when a margin is longer.


emmanuel-montoya:
thanks!

emmanuel-montoya:
I could give you a hand to merge some PRs if you want

Issue #3138, opened 2024-07-03
Memory Leak on _routesKey in RouterReportManager 


Samliangwei:
**ATTENTION: DO NOT USE THIS FIELD TO ASK SUPPORT QUESTIONS. USE THE PLATFORM CHANNELS FOR THIS. THIS SPACE IS DEDICATED ONLY FOR BUGS DESCRIPTION.**
**Fill in the template. Issues that do not respect the model will be closed.**

**Describe the bug**
I found memory leak on GetX controller when I used a memory leak detect tool. 
GC-Chain: GetxController -> state -> Page -> GetRoutePage -> RouterReportManager._routesKey
GetX version 4.6.6
**Reproduction code
NOTE: THIS IS MANDATORY, IF YOUR ISSUE DOES NOT CONTAIN IT, IT WILL BE CLOSED PRELIMINARY)**
class RouterReportManager
static final Map<Route?, List<String>> _routesKey = {};
static void _removeDependencyByRoute(Route routeName) {
    final keysToRemove = <String>[];

    _routesKey[routeName]?.forEach(keysToRemove.add);

    /// Removes `Get.create()` instances registered in `routeName`.
    if (_routesByCreate.containsKey(routeName)) {
      for (final onClose in _routesByCreate[routeName]!) {
        // assure the [DisposableInterface] instance holding a reference
        // to onClose() wasn't disposed.
        onClose();
      }
      _routesByCreate[routeName]!.clear();
      _routesByCreate.remove(routeName);
    }

    for (final element in keysToRemove) {
      final value = GetInstance().delete(key: element);
      if (value) {
        _routesKey[routeName]?.remove(element);
      }
    }
    // should add this line to prevent memery leak on _routesKey
    _routesKey.remove(routeName);

    keysToRemove.clear();
  }
example:

```dart
void main() => runApp(MaterialApp(home: Home()));

class Home extends StatelessWidget {
  final count = 0.obs;
  @override
  Widget build(context) => Scaffold(
      appBar: AppBar(title: Text("counter")),
      body: Center(
        child: Obx(() => Text("$count")),
      ),
      floatingActionButton: FloatingActionButton(
        child: Icon(Icons.add),
        onPressed: () => count.value++,
      ));
}
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Flutter Version:**
Enter the version of the Flutter you are using

**Getx Version:**
Enter the version of the Getx you are using

**Describe on which device you found the bug:**
[...]

PR #3141, opened 2024-07-04
fix memery leak: '_routesKey' holds the reference when route poped

Samliangwei:
When using the memory leak detection tool, I found a memory leak problem that is bound to occur: when the GetRoute is popped, the '_routeKey' reference in RouterReportManager holds the popped route's reference. So when this route is disposed, the reference object of the route should be removed from the "_routeKey" Map.

## Pre-launch Checklist
This PR is used to fix the ISSEU: https://github.com/jonataslaw/getx/issues/3138


jonataslaw:
Wow, that makes sense!

LGTM

PR #3320, opened 2025-02-24
Revert supported 3.22

chandrabezzo:
I see in the pubspec.yaml file supported from Flutter SDK >= 3.22. But, the code just supported on Flutter SDK >= 3.29. Need to reverted then make relevan changes in the future with updated the pubspec.yaml. I think projects that use 3.29 still limited need to support for others dependencies upgrade to the latest gradle. Better to support 3.22 for now.

### DIFF
--- a/example/lib/pages/home/presentation/views/details_view.dart
+++ b/example/lib/pages/home/presentation/views/details_view.dart
@@ -23,7 +23,7 @@ class DetailsView extends GetView<DetailsController> {
           image: DecorationImage(
             fit: BoxFit.cover,
             colorFilter: ColorFilter.mode(
-              Colors.black.withValues(alpha: 0.2),
+              Colors.black.withOpacity(0.2),
               BlendMode.darken,
             ),
             image: NetworkImage(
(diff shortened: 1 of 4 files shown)


chandrabezzo:
HI @jonataslaw thanks for response, have you check your latest master [pubspec.yaml](https://github.com/jonataslaw/getx/blob/6356d698c29ebfd68cdfa8cf8d890694989562cb/pubspec.yaml#L7-L8)? You set min version of flutter SDK support is 3.22. But, if we run with getx version `get 5.0.0-release-candidate-9.3.2` using 3.22 is impossible because have several code (method) that just supported on the latest Flutter SDK >= 3.29. 

You have two choice, update your pubspec.yaml with 3.29 as minimum, or reverted the code that just supported in 3.29 on this pull request I've created. 

I think for now better to supported projects that supported with minimum 3.22 instead of force minimum version on 3.29. Have you see this [article](https://benamorn.medium.com/today-i-learned-after-upgrading-to-flutter-3-29-0-7c7ab1a97975)? We need to support others packages or plugins to use 3.29 as our minimum Flutter SDK.

PR #3361, opened 2025-06-06
feat(controllers): enhance documentation, add lifecycle callbacks, an…

Aniketkhote:
- Improve documentation for all controller classes with usage examples
- Add error handling in FullLifeCycleMixin callbacks
- Implement onMemoryPressure and onAccessibilityChanged callbacks
- Add proper ScrollController disposal in ScrollMixin
- Enhance type safety and error reporting
- Update documentation with clear method descriptions and examples
- Fix potential memory leaks in lifecycle management

### DIFF
--- a/lib/get_state_manager/src/simple/get_controllers.dart
+++ b/lib/get_state_manager/src/simple/get_controllers.dart
@@ -15,7 +15,7 @@ import 'list_notifier.dart';
 /// ```dart
 /// class CounterController extends GetxController {
 ///   var count = 0;
-///   
+///
 ///   void increment() {
 ///     count++;
 ///     update(); // Triggers UI update
@@ -34,28 +34,16 @@ abstract class GetxController extends ListNotifier with GetLifeCycleMixin {
   /// - [ids]: Optional list of widget IDs to update. If null, updates all widgets.
   /// - [condition]: If false, the update will be skipped.
   ///
-  /// Throws:
-  /// - [FlutterError] if the controller is not properly initialized.
   void update([List<Object>? ids, bool condition = true]) {
     if (!condition) {
       return;
     }
-    try {
-      if (ids == null) {
-        refresh();
-      } else {
-        for (final id in ids) {
-          refreshGroup(id);
-        }
+    if (ids == null) {
+      refresh();
+    } else {
+      for (final id in ids) {
+        refreshGroup(id);
       }
-    } catch (error, stackTrace) {
-      FlutterError.reportError(FlutterErrorDetails(
-        exception: error,
-        stack: stackTrace,
-        library: 'get_state_manager',
-        context: ErrorDescription('while updating controller state'),
-      ));
-      rethrow;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


lib/get_state_manager/src/simple/get_controllers.dart
```diff
@@ -109,28 +273,84 @@ mixin FullLifeCycleMixin on FullLifeCycleController {
...
-      case AppLifecycleState.paused:
-        onPaused();
-        break;
-      case AppLifecycleState.detached:
-        onDetached();
-        break;
-      case AppLifecycleState.hidden:
-        onHidden();
-        break;
+    try {
```
jonataslaw, line 276: Here the error would not be coming from the library code, but from within some function in the user code. This could obfuscate the debugging. Unless we use something like runZoned (which would be overkill), this is unnecessary in my opinion.

lib/get_state_manager/src/simple/get_controllers.dart
```diff
@@ -109,28 +273,84 @@ mixin FullLifeCycleMixin on FullLifeCycleController {
...
+      }
+    } catch (error, stackTrace) {
+      _reportError('Error in didChangeAppLifecycleState', error, stackTrace);
+    }
+  }
+
+  @override
+  void didHaveMemoryPressure() {
+    super.didHaveMemoryPressure();
+    try {
```
jonataslaw, line 302: Same here

lib/get_state_manager/src/simple/get_controllers.dart
```diff
@@ -109,28 +273,84 @@ mixin FullLifeCycleMixin on FullLifeCycleController {
...
+      onMemoryPressure();
+    } catch (error, stackTrace) {
+      _reportError('Error in didHaveMemoryPressure', error, stackTrace);
+    }
+  }
+
+  @override
+  void didChangeAccessibilityFeatures() {
+    super.didChangeAccessibilityFeatures();
+    try {
```
jonataslaw, line 312: and here

jonataslaw (changes requested):
Thank you very much for this PR. The documentation is generally well written, but I have some concerns regarding the try/catch blocks.

The first one is meant to handle a case where `update()` is called before the controller has been initialized. However, this is impossible, since the constructor itself initializes the controller. In other words, by the time we call `controller.update()`, the controller is already initialized. This makes the try/catch both unnecessary and a source of extra overhead.

Second, the lifecycle methods are meant to be overridden by the user. If they override without calling `super`, the try/catch will never execute; if they override and call `super`, they’ll simply run an empty block (or Flutter’s default code). In my opinion, this try/catch is also unnecessary and can actually confuse debugging: if an error is thrown, the stack trace will point to the library code rather than the function that actually caused the error.

Aniketkhote:
I've addressed the feedback with the following changes:

Removed try-catch blocks from lifecycle methods to align with Flutter's idiomatic error handling, allowing better debugging and letting global handlers catch issues.

Simplified the update() method by removing redundant error handling—it's safe as the controller is already initialized.

Kept documentation improvements for better developer clarity.

These changes follow Flutter best practices, giving developers more control while improving maintainability.

jonataslaw:
Thank you!
LGTM!

fisforfaheem:
any new updates??

PR #3396, opened 2025-08-31
Sealed Classes & Status Handling Refactor

Aniketkhote:
### ✨ New Features

* **Sealed Classes for Status Handling**

  * Refactored `GetStatus` to use Dart’s sealed classes for improved type safety and pattern matching
  * Added **exhaustive pattern matching** with the `match` method
  * Added **`when` method** for side-effect-based status handling
  * Added **`mapSuccess`** for transforming success values
  * Added **`dataOrNull`** and **`errorOrNull`** extensions for safer data access

### 🔧 Improvements

* **Type Safety**: Stronger typing across state management APIs
* **Documentation**: Comprehensive docs for all public APIs
* **Testing**: Extended test coverage for new sealed class implementation
* **Performance**: Optimized state updates with efficient equality checks

### ⚠️ Breaking Changes

* `GetStatus` is now a **sealed class** with a private constructor
* `ErrorStatus` now accepts **only one type parameter**
* Custom status implementations must now **extend predefined status classes**
* `isCustom` getter only returns `true` for `CustomStatus` instances

#### 1. Pattern Matching

Replace `if-else` chains with `match`:

```dart
// Before
if (status.isLoading) {
  return LoadingWidget();
} else if (status.isError) {
  return ErrorWidget(status.errorMessage);
} else {
  return SuccessWidget(status.data);
}

// After
return status.match(
  loading: () => LoadingWidget(),
  error: (error) => ErrorWidget(error?.toString() ?? 'Unknown error'),
  success: (data) => SuccessWidget(data),
  empty: () => EmptyWidget(),
  custom: () => CustomWidget(),
);
```

#### 2. Error Status

Update usages of `ErrorStatus` with **two type parameters** → now only **one**:

```dart
// Before
ErrorStatus<MyData, Exception> status;

// After
ErrorStatus<MyData> status;
```

#### 3. Custom Status

Custom implementations should now **extend predefined status classes**.

---

✅ **This PR makes state handling safer, more expressive, and future-proof.**


lib/get_state_manager/src/rx_flutter/get_status.dart
```diff
@@ -0,0 +1,147 @@
...
+/// - [CustomStatus]: A custom status for specialized use cases
+sealed class GetStatus<T> with Equality {
+  const GetStatus._(); // Private constructor for the sealed class
+
+  // Factory constructors for each variant
+  const factory GetStatus.loading() = LoadingStatus<T>._;
+  const factory GetStatus.error([Object? error]) = ErrorStatus<T>._;
+  const factory GetStatus.empty() = EmptyStatus<T>._;
+  const factory GetStatus.success(T data) = SuccessStatus<T>._;
+  const factory GetStatus.custom() = CustomStatus<T>._;
```
hashirshoaeb, line 20: API does not match the previous options. 

```dart
  const factory RxStatus.loading() = RxLoading;
  const factory RxStatus.loadingMore() = RxLoadingMore;
  const factory RxStatus.success() = RxSuccess;
  const factory RxStatus.error([String? message]) = RxError;
  const factory RxStatus.empty() = RxEmpty;
  ``` 
  
  loadmore is missing. and custom is new. 
  error accept string not object.

File: lib/get_state_manager/src/rx_flutter/get_status.dart [hunk 2]
```diff
@@ -0,0 +1,147 @@
...
+  T? get dataOrNull => switch (this) {
+    SuccessStatus(data: var d) => d,
+    _ => null,
+  };
+
+  /// Returns the error if this is an [ErrorStatus], otherwise returns null
+  Object? get errorOrNull => switch (this) {
+    ErrorStatus(error: var e) => e,
+    _ => null,
+  };
```
hashirshoaeb, line 49: I this is not needed i guess

lib/get_state_manager/src/rx_flutter/get_status.dart
```diff
@@ -0,0 +1,147 @@
...
+    R Function(Object? error)? error,
+    R Function()? empty,
+    R Function(T data)? success,
+    R Function()? custom,
+  }) {
+    return switch (this) {
+      LoadingStatus() =>
+        loading?.call() ?? (throw StateError('No handler for loading state')),
+      ErrorStatus(error: var e) =>
+        error?.call(e) ?? (throw StateError('No handler for error state')),
```
hashirshoaeb, line 138: i like the way you think. but i feel this is too much apis.

Aniketkhote: So, are you suggesting we should just keep the existing APIs as they are and mainly add sealed, without making bigger changes?

